PR #37, opened 2014-12-30
First draft of algorithms for the specification.

mfoltzgoogle:
This PR is a first draft of a set of algorithms for the specification.  The specification style is loosely based on that in [Service Workers](http://www.w3.org/TR/service-workers/).

There are several open issues noted.  Some of them can possibly be closed before merging this; some of them may remain open issues that need further discussion.

The Makefile in the current repository does not correctly extract and apply xrefs; it seems that the definitions are missing.  I commented out the parts of the Makefile that use xrefs so I could use tidy.  Please advise if there are missing references I can pull in from the current repo.

Thanks!
Mark


drott:
As discussed, merged pull request #38 instead.

PR #43, opened 2015-01-21
Addressing Anssi's feedback, expanding the noted items

drott:
* APIs for interacting with Presentation.. => "API"
* opening browsing context
* close, send => own algorithms
* send TODO moved
* session property open-issue, still required?
* AvailableChange event reorg'ed
* Idioms => uppercase
* Definition of "presentation"
* session identifier => add "alphanumeric"
* Availability updating algorithm isolated

* Organizing section of onavailable change
** Define monitoring algorithm
** Define when to start and cancel monitoring

* Event handlers wording
* Replace "opener page" => opening browsing context
* removed .XXX style
* Padding before headings

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -65,6 +65,11 @@
     }
 
     code { color: orangered; }
+    .XXX { background: white; border: solid #78AB46; padding: 0.5em; margin: 1em 0; }
+    table { border-collapse: collapse; border-style: hidden hidden none hidden; }
+    table thead, table tbody { border-bottom: solid; }
+    table td, table th { border-left: solid; border-right: solid; border-bottom: solid thin; vertical-align: top; padding: 0.2em; }
+    dfn { font-weight: bolder; font-style: normal; }
     </style>
   </head>
   <body>
@@ -255,7 +260,7 @@
       devices through any of the above means.
     </p>
     <h3 id="use-cases">
-      Use Cases
+      Use cases
     </h3>
     <h4>
       Presentations
@@ -275,7 +280,7 @@
       <b>Requirements:</b> R1, R3, R4, R5, R7
     </p>
     <h4>
-      <a id="video-sharing">Video and Image Sharing</a>
+      <a id="video-sharing">Video and image sharing</a>
     </h4>
     <p>
       Using an online video or image sharing service, a user would like to show
@@ -306,7 +311,7 @@
       <b>Requirements:</b> R1, R3, R4, R5, R7
     </p>
     <h4>
-      Media Flinging to Multiple Screens
+      Media flinging to multiple screens
     </h4>Alice enters a video sharing site using a browser on her tablet. Next,
     Alice picks her favorite video from the site, and the video starts to play
     on her tablet. While the video is playing Alice clicks a button "Share on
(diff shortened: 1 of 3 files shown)


drott:
Preview at https://rawgit.com/drott/presentation-api/anssi_reorg/index.html

Overview.src.html
```diff
@@ -750,200 +755,246 @@ <h3 id="use-cases">
...
-      "#previous-version">previous version of the Presentation API</a>. This
-      section describes the interfaces to the extend discussed in the Second
-      Screen Presentation Community Group. Readers are encouraged to consult
-      the <a href="#example">Example</a> and <a href=
-      "#algorithms">Algorithms</a> sections together with this section for a
-      more complete understanding of the technical parts of this specification.
-    </p>
     <h3>
-      <code>NavigatorPresentation</code>
+      Common Idioms
```
anssiko: Lowercase "idioms" for consistency :-)

Overview.src.html
```diff
@@ -750,200 +755,246 @@ <h3 id="use-cases">
...
+    <p>
+      An <dfn>opening browsing context</dfn> is a <span data-anolis-spec=
+      "w3c-html">browsing context</span> from which calls are made to
+      <span>startSession</span>, <span>joinSession</span>, in other words from
+      which <span title="concept-presentation-session">presentation
+      sessions</span> are initiated.
+    </p>
     <h3>
-      <code>AvailableChangeEvent</code>
+      Common Definitions
```
anssiko: Lowercase "Definitions" for consistency :-)

anssiko:
LGTM.

mfoltzgoogle:
Overview.src.html:

L669 Instead of ‘requesting page’, can we say

> The opening browsing context can then communicate with the presentation as if the user had manually connected to it via startSession.

or just

> The page can then communicate with the presentation as if the user had manually connected to it via startSession.

L787 I might say:

> A presentation is an active connection between a user agent and a presentation display for displaying web content on the latter at the request of the former.

L793 I might say

> A presentation session is an object relating an opening browsing context to its presentation display and enabling two-messaging between them. Each object has a presentation session state and a presentation session identifier to distinguish its session from other presentation sessions.

L801 I might say: 

> An opening browsing context is a browsing context that has initiated or resumed a presentation session by calling startSession or joinSession.

L986
> (drott) What would be a good definition of the session attribute? Do we want to keep it? It's somewhat hard to define, when it should contain a value, and when it should be empty - and it's not compatible with multiple calls to startSession, for example.

The purpose of the session attribute was to provide the presented page with access to the PresentationSession without allowing a potential race condition between the registration of an event handler and the firing of an event, or complicating the browser implementation to fix this case.  It should be null in an opening context and non-null in a presented document.

I think we should reconsider this design in light of our use of Promises in other parts of the spec, as well as our intention to allow multiple opener pages to communicate with a given presentation.

L1343 I feel this sentence (or some improved version) should be put in the overview (as modified below) to replace “In the following...” to try to tie together the three algorithms.  But I don’t feel strongly.

> The user agent must keep a list of available presentation displays.  According to the number of event handlers for onavailablechange, the user agent must also keep the list up to date by running the steps in Monitoring the list of available presentation displays.

mfoltzgoogle:
For the last suggestion, I was specifically suggesting the text be inserted under heading 6.4.4 "The onavailablechange EventHandler."

Issue #46, opened 2015-01-30
Define a bi-directional data channel between opening and presenting browsing contexts


drott:
http://w3c.github.io/presentation-api/#algorithm-post-message

Consider additional binary datatypes? https://bugzilla.mozilla.org/show_bug.cgi?id=1080474#c22

mfoltzgoogle:
Candidates, similar interfaces, are found in the [WHATWG HTML spec's `MessageEvent` section] (https://html.spec.whatwg.org/multipage/comms.html#the-messageevent-interfaces) or [WebRTC spec for `RPCDataChannel`](http://w3c.github.io/webrtc-pc/#rtcdatachannel)

mfoltzgoogle:
Mozilla currently supports [this interface for messaging](https://bugzilla.mozilla.org/attachment.cgi?id=8527264&action=diff#a/dom/webidl/PresentationSocketChannel.webidl_sec1):

```
interface PresentationSession : EventTarget {
  <snip>
  readonly attribute PresentationSocketChannel channel;
};
```

```
interface PresentationSocketChannel : EventTarget {
17	  [Throws]
18	  void send (DOMString data);
19	  [Throws]
20	  void send (Blob data);
21	  [Throws]
22	  void send (ArrayBuffer data);
23	  [Throws]
24	  void send (ArrayBufferView data);
25	
26	  void close ();
27	
28	  attribute PresentationBinaryType binaryType;
29	  readonly attribute unsigned long bufferedAmount;
30	
31	  const unsigned short CONNECTING = 0;
32	  const unsigned short OPEN = 1;
33	  const unsigned short CLOSING = 2;
34	  const unsigned short CLOSED = 3;
35	  readonly attribute unsigned short readyState;
36	
37	  attribute EventHandler onmessage;
38	  attribute EventHandler onopen;
39	  attribute EventHandler onerror;
40	  attribute EventHandler onclose;
41	};
```

I like that the API supports multiple data types, but it seems to assume the underlying connection technology has the same semantics as [WebSockets](http://www.w3.org/TR/websockets/).

My questions for the Mozilla folks:

(1) What does it mean for the PresentationSession to be connected but the associated channel to be CLOSED?  Similarly what does it mean for the session to be disconnected but the channel != CLOSED?

(2) What if the connection technology between the user agents is connection-less (i.e. based on UDP or a push messaging service)?  What would the readyState or state transitions mean?

[...]

PR #62, opened 2015-03-06
First version of postMessage and onmessage algorithms

drott:
Modeled after WebSockets behavior.

Addresses issue #46.

TODO: Fix references to ArrayBuffer, Blob, etc.
TODO: Discuss whether we need "binaryType"
TODO: Check _trusted_ event property. 
TODO: Define presentation origin somewhere.


drott:
Rawgit HTML version here: https://rawgit.com/drott/presentation-api/postmessage/index.html#posting-a-message-through-presentationsession

mfoltzgoogle:
Thanks for putting this together, comments:

Indent the following to match the rest of the definition.   Also the first instance of 'postMessage' is bold while other declarations are not.

```
// Communication
attribute BinaryType binaryType;
EventHandler onmessage;
void postMessage (DOMString message);
void postMessage (Blob data);
void postMessage (ArrayBuffer data);
void postMessage (ArrayBufferView data);
```

> the data contents of a transmission between two browsing contexts

Does this mean a single invocation of `postMessage()`?  I think it would be permissible for the implementation to fragment transmission (e.g. to meet the requirements of a network transport) of a single message.  Can it be clarified?

> Let presentation message format be the format of that data, one of Text and Binary.

I think of this as a type, not a format; the text data could be sent in a binary format if the implementation chooses.  So maybe: "Let *presentation message type* be the type of that data, one of `text` or `binary`."

> (Set to Binary if data is one of ArrayBuffer or Blob. Set to Text if data is of type DOMString)

This would be clearer as its own step.

> to the remote side, i.e. from opening browsing context to presentation browsing context or vice versa

Maybe just "to the destination browsing context"?  Or have a step that defines the destination context for the data?

>the URL that the opening browsing context and the presenting browsing context have in common.

What if they have no URL in common?

> If the presentation message format is Text, then initialize event's data attribute to the contents of  presentation message data.

...as a DOMString?

drott:
> TODO: Fix references to ArrayBuffer, Blob, etc. …

Added, drott's xref repo updated to include TypedArray spec.

> TODO: Discuss whether we need "binaryType"

Left in for now to emulate WebSocket spec.

> TODO: Check trusted event property.

Reference to trusted event added, isTrusted is true since it's a UA generated event.

> TODO: Define presentation origin somewhere.

Issue #63 created.

> Does this mean a single invocation of postMessage()?
> I think it would be permissible for the implementation to fragment
> transmission (e.g. to meet the requirements of a network transport)
> of a single message. Can it be clarified?

Tried to clarify this by specifying a "data payload" and an implementation specific transport mechanism.

>> presentation message format

> I think of this as a type, not a format;

Changed.

>> (Set to Binary if data is one of ArrayBuffer or Blob. Set to Text if data is of type DOMString)

> This would be clearer as its own step.

Moved to own step.

> Maybe just "to the destination browsing context"? Or have a step that defines the destination context for the data?

Defined destination browsing context in separate step.

> What if they have no URL in common?

See above, created issue #63 to discuss origin relationship.

> ...as a DOMString?

Added mention of type here.

mfoltzgoogle:
LGTM with a few minor fixes:

L1025: format => type
L1052-1070: 
format => type
Binary -> `binary`
Text -> `text`

drott:
Thanks for the thorough review.

Issue #42, opened 2015-01-19
Make the Example section more concise


anssiko:
As we make progress with the spec and the normative sections begin to be self-contained, we may want to carve out the now very extensive Requirements and Example sections into its own "Use Cases and Requirements (and Examples)" document we can evolve in tandem with the API specification at a later stage, referenced from the API spec. Perhaps leave one concise example in the spec.

anssiko:
After @mfoltzgoogle's good changes to the spec to map GitHub issues to the spec, I think we should keep the use cases and requirements in the spec, since they provide good background information, especially for new people reviewing the specification who have not been in the group (or in the Community Group) from its start.

However, I feel the Example section is still too elaborate currently, and requires more work as we progress along the Rec Track. To that end, I'll rescope this issue and rename to "Make the Example section more concise".

mfoltzgoogle:
We could include a smaller code snippets showing how to start/connect to a presentation, and eliminate the prose that has become redundant with the algorithm descriptions in the API section (after addressing the open issued described).

anssiko:
I agree with @mfoltzgoogle in that we should include smaller code snippers in the Example section, and eliminate the prose as it becomes redundant. Optimally, the API is self-documenting to the extend no extensive developer documentation is needed to understand otherwise well-documented example code.

I will unassign myself from this issue, since I think this would be a good first issue for someone new to take who'd like to contribute to the Example section directly.

All - Feel free to assign this issue to yourself, if you'd like to help with this issue in the near future.

Writing code examples against the API and keeping them up to date as the API evolves is a good practise as it prevents surprises with developer ergonomics, helps identify functional issues and spec bugs before the actual implementation feedback is received.

louaybassbouss:
@anssiko you can assign this task to me I am this week on business trip I will work on it on Monday next week

anssiko:
@louaybassbouss Thanks for volunteering. You should be able to click the "assign yourself" link visible on the desktop version of the site (not visible in the mobile version, apparently).

louaybassbouss:
@anssiko it seems that I don't have the rights to assign myself an existing issue

louaybassbouss:
This is the first try for the Examples section. If you receive this as mail notification please open it in Github by clicking on the link at the end of the mail for better formatting.

# Examples

[...]

PR #64, opened 2015-03-16
Make the Example section more concise  (Issue #42)

louaybassbouss:
This PR contains new draft of the example section as discussed in issue #42

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -566,16 +566,13 @@
   var presUrl = "http://example.com/presentation.html";
   // create random presId 
   var presId = Math.random().toFixed(6).substr(2);
-  navigator.presentation.startSession(presUrl, presId).then(function(newSession){
-    // a new session is started successfully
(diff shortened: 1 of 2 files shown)


louaybassbouss:
Hi Dominik, I just pushed a new version. 

BTW is there another code formatting style with better code highlighting like in some other W3C Specs.

BR

drott:
LGTM, thanks for the quick update.

drott:
> BTW is there another code formatting style with better code highlighting like in some other W3C Specs.

Any ideas about that @anssiko ? If nothing else, we could link ```highlight.js``` for example.

anssiko:
Anolis does not have built-in support for syntax highlighting, while some other spec authoring tools such as ReSpec and Bikeshed do have it.

If we'd enable highlighting it'd be better done at build time. There's a Node.js module for `highlight.js`, so perhaps that is doable with little effort.

PR #75, opened 2015-04-09
A link to WG's charter

hongkicha:
Hi, all. At the **Status of This Document** Section, I specified our WG’s [charter](http://www.w3.org/2014/secondscreen/charter.html). <br>A WG's charter is the first document when a non-member gets to read before joining a WG. I’m aware that our charter is well-specified in the Work Mode. For non-members, however, I believe that the fastest and easiest way to access the WG charter is to add it in the beginning of the working document. (just like they did for HTML5 recommendation)
Thank you for your consideration.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -175,7 +175,10 @@
       <p>
         This document was published by the <a href=
         "http://www.w3.org/2014/secondscreen/">Second Screen Presentation
-        Working Group</a> as an Editor's Draft. If you wish to make comments
+        Working Group</a> as an Editor's Draft. The Working Group actively
+        pursues its activities within the bounds of the <a href=
+        "http://www.w3.org/2014/secondscreen/charter.html">W3C Second Screen
+        Presentation Working Group charter</a>. If you wish to make comments
         regarding this document, please send them to <a href=
         "mailto:public-secondscreen@w3.org">public-secondscreen@w3.org</a>
         (<a href=
--- a/index.html
+++ b/index.html
@@ -78,8 +78,8 @@
       <h1>
         Presentation API
       </h1>
-      <h2 class="no-num no-toc" id="editor's-draft-6-april-2015">
-        Editor's Draft 6 April 2015
+      <h2 class="no-num no-toc" id="editor's-draft-9-april-2015">
+        Editor's Draft 9 April 2015
       </h2>
       <dl>
         <dt>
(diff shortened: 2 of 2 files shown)


mfoltzgoogle:
@anssiko, @tidoust What do you think of this request?

My only concern is that the charter may change over time, and whether we should link the most recent version, or to the published version that governs the group's current work.

tidoust:
@mfoltzgoogle The "Status of This Document" section is the only place in the document that has some temporal aspect to it. It's also a section that will change basically each time we publish the document. What it describes is *current at the time of publication*, whereas the rest of the specification will hopefully remain valid over time.

Also, the charter will change when and if it is renewed, but the previous charter will be updated in place to alert the reader that it is no longer valid and to link to the new version.

We may include a link to the charter in short, although It might be worth clarifying that the charter is valid when the document is published, for instance with something as simple as:

"This document was published by the *Second Screen Presentation Working Group* as an Editor's, operating within the bounds of its *charter*".

I do not know how many potential readers know or care about what the charter says, unless they want to get involved. I note that HTML5 is an exception to the rule for many things, this included. In practice, I am not aware of other specs that link directly to the group's charter from their Status of This Document sections. Readers willing to know more about who produced the spec and within what bounds can follow the link to the group's home page, which is less dry than the charter and which links to it.

In the end, I guess I do not see a compelling reason to introduce another link in the Status of This Document section, which has a number of them already, but that's just my take on it.

mfoltzgoogle:
Per Francois' comment, it seems the preferred entry point for non-members is the [Second Screen Working Group's home page](http://www.w3.org/2014/secondscreen/), where the charter is prominently linked.

And since the charter will link to previous versions, I also do not see a need to tie together a specific revision of the spec with a specific revision of the charter, as non-members will be able to find it themselves.

It appears the HTML5 spec is using the charter to help clarify the relationship between the HTML WG and the activities of the WHATWG.  I don't think we have the same situation here.

anssiko:
I'd recommend we leave SoTD as is for the reasons noted above.

anssiko:
I will close this PR since there appears to be an agreement no changes to the SoTD needed.

hongkicha:
@anssiko @mfoltzgoogle @tidoust  Thank you very much for your clarification and sorry for the long absence. Thanks to you all, now I got to know more what a charter means to W3C. Cheers.

Issue #26, opened 2014-11-04
Allow page to designate a default presentation URL


mfoltzgoogle:
Some user agents may wish to allow the user to initiate presentation from the browser instead of by triggering it via an action on the page.  This is analogous to choosing "Print" from the browser menu instead of calling window.print().

In this scenario, the page should be able to designate a default presentation URL that will be shown when the user initiates presentation via the browser.  This issue tracks the work to define the specification and the user-agent behavior around this URL.

avayvod:
Hey everyone,

as I'd like to implement this bit of the API in Chrome, I'd like to try to get a consensus on the thread [1] and add the result to the spec if possible. As I see it, the viable extra specification would consist of two parts:

1) <link rel="presentation-url" href="..."/> and <link rel="presentation-id" href="..."/> to allow the page to define and change the default presentation URL and id to be used by the UA in the UA-initiated-session scenario; basically any external user action that might start a presentation (e.g. clicking the present button in UA toolbar, tapping on an NFC tag, etc) would be equivalent to the page calling startSession(presentationUrl, presentationId) with the values from the <link> tags.

2) Upon successful connection and session initialization, |navigator.presentation.defaultsessionstart| event would be fired by the UA to all pages with the matching <presentation-url, presentation-id> pair containing the PresentationSession object allowing the page to communicate with the initiated session:

[
    Constructor(DOMString type, optional DefaultSessionStartEventInit eventInitDict)
] interface DefaultSessionStartEvent : Event {
    readonly attribute PresentationSession session;
};

Thoughts?

Thanks,
Anton.

[1] This was discussed in November last year, see http://lists.w3.org/Archives/Public/public-webscreens/2014Nov/0007.html

avayvod:
From mfoltz:

> The |defaultsessionstart| event should only be fired on the top level frame/document in a given page.
> If there are multiple <frames> declaring default presentations, there is no easy way for the user 
> agent to disambiguate the user intent to initiate presentation.

I think this is mostly a UI issue we should leave for the UAs to solve. The spec may mention it but leave it to the implementors to decide what to do (e.g. one could show a list of DPUs available and ask user to pick one).

> If the |defaultsessionstart| event handler is attached to the |navigator| object, how do we ensure
> that only the frame that declared the <link> tags gets access the session?

Well, there might me more than one frame / page that declared the same DPU. It seems right to notify all of them; the page may ignore the event if its document is not in focus.

avayvod:
[...]

PR #77, opened 2015-04-15
Added defaultsessionstart and default-presentation <link> to the spec

avayvod:
See https://github.com/w3c/presentation-api/issues/26 for details.

I also fixed the <section> structure of the document which was misaligned and updated xrefs/presentation.json (not sure if I had to do that or not).

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -392,8 +392,9 @@
               Launching presentation
             </dt>
             <dd>
-              The UA must provide a means of start sending content to the
-              secondary screen.
+              The UA must provide a way to start sending content to the
+              secondary screen. This may occur at the request of the page or at
+              the request of the UA.
             </dd>
             <dt>
               Resuming presentation
@@ -601,6 +602,25 @@
       </section>
       <section>
         <h3>
+          Handling an event for a UA initiated presentation session example
+        </h3>
+        <pre class="example">
+&lt;!-- controller.html --&gt;
+&lt;head&gt;
+  &lt;!-- the link element with rel='default-presentation' allows the page to specify --&gt;
+  &lt;!-- the presentation URL and id for when the UA initiates a presentation session --&gt;
+  &lt;link href="http://example.com/presentation.html" rel="default-presentation" id="ABCD1234" &gt;
+&lt;/head&gt;
+&lt;script&gt;
+  navigator.presentation.ondefaultsessionstart = function (evt) {
+    setSession(evt.session);
+  };
+&lt;/script&gt;
+
+</pre>
+      </section>
+      <section>
+        <h3>
           Monitor session's state and exchange data example
         </h3>
         <pre class="example">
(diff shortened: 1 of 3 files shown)


avayvod:
See http://avayvod.github.io/ for the new version of the spec.

anssiko:
Thanks for the PR! Some comments below:

* For "Note: ..." please use `<p class="note">...</p>`

* The new link type "default-presentation" is used in the example, but it is not defined in the normative prose. Please add a new section for this. You can refer to the Link types section [1] in the HTML spec for some inspiration.

* In the "6.6 Interface DefaultSessionStart" section [2] it is said: 

> An event named defaultsessionstart is fired when the UA initiates a presentation on behalf of the page.

We should specify at which stage of the "establish a presentation connection" algorithm [3] the event is fired.

* Could we fire the `defaultsessionstart` event whenever a new session is started, regardless of whether it is UA initiated or a result of `startSession()` or `joinSession()` invocation?

[1] http://www.w3.org/TR/html51/semantics.html#linkTypes
[2] http://avayvod.github.io/#interface-defaultsessionstart
[3] http://avayvod.github.io/#establish-a-presentation-connection

mfoltzgoogle:
@avayvod, comments on the PR.

Overview.src.html:

L397: Instead of adding a Note:, I might phrase as:
> The UA must provide a way to start sending content to the secondary screen.  This may occur at the request of the page or at the request of the UA.

L605: Maybe:
> Handling an event for a UA initiated presentation session example

L611:
> ...allows the page to...

I suggest defining an algorithm for starting a browser initiated presentation session, that initializes session and updates the session database `D`, then chains to 6.4.3 (*Establishing a Presentation Session Connection*).

Do we need to update  *Monitoring the list of available presentation displays* to take into account the default presentation URL?

mfoltzgoogle:
@anssiko, in response to 

> Could we fire the defaultsessionstart event whenever a new session is started, regardless of whether it is UA initiated or a result of startSession() or joinSession() invocation?

We could fire this event in all cases, but the caller would need to handle the duplicate session returned by both `defaultsessionstart` and from resolving the `Promise`s returned by `startSession()` and `joinSession()`.  Do you see an advantage to that?

anssiko:
@mfoltzgoogle Sorry if this was pending for my response, I missed this. Please feel free to rebase and merge.

avayvod:
I believe we could have a generic onsessionstart event then and abandon returning a Promise from start/joinSession resulting in a more unified API and smaller API surface. Not sure it would be convenient for the page because it might have to track what request is being resolved with the new session passed in onstart. Although the page might not care much.

Having a Promise returned for startSession/joinSession is better for error handling though - the page makes a request and either gets a session or gets an error. There's no issue with the timing of when event listener is attached. If we use an event handler to pass the result of startSession/joinSession to the page, we'd have to add onsessionfail, for example.

I think a Promise based API for startSession/joinSession is better.

avayvod:
I'm still working on addressing the comments (slowly). Will try to upload a new pull request today.

avayvod:
I moved the preview of the changed spec to avayvod.github.io/defaultsessionstart.html

mfoltzgoogle:
Pull request looks good; merged.  I will check the issues and close if there is one that is relevant to this spec change.

PR #85, opened 2015-05-13
Fixed the structure of <section> tags in the spec HTML

avayvod:
Mostly some formatting changes to fix the structure of the spec (and ease modifying it).

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -6,14 +6,14 @@
       Presentation API
     </title>
     <style>
-@media print {
-     [data-anolis-spec]::after { content:"[" attr(data-anolis-spec) "]"; font-size:.6em; vertical-align:super; text-transform:uppercase }
+    @media print {
+      [data-anolis-spec]::after { content:"[" attr(data-anolis-spec) "]"; font-size:.6em; vertical-align:super; text-transform:uppercase }
     }
     </style>
     <link rel="stylesheet" type="text/css" href=
     "http://www.w3.org/StyleSheets/TR/W3C-[STATUS]">
     <style type="text/css">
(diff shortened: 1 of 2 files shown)


avayvod:
See the fixed version at http://avayvod.github.io/index.html (note that NavigatorPresentation and AvailableChangeEvent interfaces are under the API item and availablechange related sections are under availablechange event item).

mfoltzgoogle:
@avayvod, the diff in GitHub shows a lot of whitespace changes, make it hard to view any actual changes.

Can you make sure you are using the latest version of tidy-html5 (not regular tidy) as linked here?
https://www.w3.org/wiki/Second_Screen/Work_Mode

If so, then it seems the spec document is not properly formatted to begin with, I can submit a patch to reformat it and then you can rebase your patch on that.

avayvod:
Hi Mark,

I've updated tidy-html5 per the documentation, eliminated some accidental whitespace changes and uploaded the new patch. PTAL.

There're still some small whitespace changes for indents or removing blank lines in the end of the preformatted scripts. The main change starts at the line 988 where one <section> tag was closed too early causing the second API interface section be on the same level as the general "API" section. I indented the following html to reflect the correct section structure.

I've uploaded a diff of the raw text of the index.html with and without the patch and uploaded it here: http://avayvod.github.io/diff-with-patch-without-patch

As you can see there, the user visible change is only renumbering of the TOC and section titles (and two extra blank lines deleted).

There is a huge chunk in the spec file that's commented out which I deleted - I think if there's something valuable, we can always restore it using git. Right now it just clutters the spec source.

WDYT?

mfoltzgoogle:
LGTM, merged.

Issue #63, opened 2015-03-09
Define (cross) origin relationship between controller and presentation


drott:
Do we require that the opener and presenting page share an origin? Can we allow CORS?

schien:
This has also been discussed in Mozilla [dev-webapi](https://groups.google.com/forum/#!topic/mozilla.dev.webapi/C5D8hEZmCLk). We can learn from `window.postMessage()` and add additional `targetOrigin` parameter in `PresentationSession.send()`.

```
void send( (DOMString or Blob or ArrayBuffer or ArrayBufferView) data, DOMString targetOrigin);
```
To send the message to the target regardless of origin, set the target origin to `"*"`. To restrict the message to same-origin targets only, without needing to explicitly state the origin, set the target origin to `"/"`.

anssiko:
Thanks for the proposal, added to the F2F agenda.

anssiko:
@schien Could you outline your proposal at the F2F and share your latest thinking? It seems you've hashed this out on dev-webapi somewhat. Also, provide couple of good use cases that require cross-origin for background.

The `targetOrigin` attribute seems like a feasible solution the group should explore further.

Actually, the `origin` attribute in the spec [1] would be redundant information without an ability to do cross-origin (we can get that information via `window.location.origin`). The prose around `origin` should be clarified once we resolve this issue.

@mfoltzgoogle What is Google's take on this issue?

[1] http://w3c.github.io/presentation-api/#receiving-a-message-through-presentationsession

mfoltzgoogle:
For now, I don't really see the cross-origin restriction as applying to the Presentation API.   The messaging channel is analogous to the device-to-device communication implemented in `RTCDataChannel` [1]; it's just a data channel and isn't responsible for authenticating either party.

If the presenting page wishes to authenticate the identity of the presentation page (or vice versa) the initial handshake between the pages could include the exchange of tokens that can be used to prove that identity.  This minimizes the level of trust required in both the user agents rendering each side and the security requirements of the network transport.

However, let's assume that we trust both user agents and the network transport, so the origin of either party is conveyed truthfully.  

- How would one side know to update the `targetOrigin` for messages as the other side navigated to a different origin and rejoined the same `PresentationSession`?

- If pages from different origins A and B wished to join the same PresentationSession simultaneously, what target origin would that presentation use to communicate?

This is a good topic to discuss further at the F2F; a more fully defined proposal for cross-origin messaging would be a good starting point.

[...]

PR #88, opened 2015-05-19
Fixed typos and removed the 'origin' extraction step from the send() algorithm

avayvod:
A demo link is at http://avayvod.github.io/nits.html
The removal of the unused 'origin' step from the PresentationSession.send() was discussed at the F2F as part of the Issue #63


mfoltzgoogle:
You can remove the issue reference from the document; if we do add something CORS related for packaged Web apps I think that would happen under a separate issue / feature request.

Issue #81, opened 2015-04-27
Rethinking availability monitoring


anssiko:
The current events-based API for availability monitoring relies on the following side-effect: adding/removing a listener for the `availablechange` event starts/stops discovery. This is discouraged (see the historical discussion [1], also implementability concerns [2]).

A promise-based API is proposed as the solution [2][3].

[1] https://lists.w3.org/Archives/Public/public-device-apis/2011Sep/0025.html
[2] https://lists.w3.org/Archives/Public/public-secondscreen/2015Apr/0079.html
[3] https://lists.w3.org/Archives/Public/public-secondscreen/2015Apr/0068.html

obeletski:
Hi All,

First of all, I agree with Anssi, that availability monitoring could be improved and because of multiple reasons. One of the reasons is that current API design makes it very simple for web developers to start continuous monitoring and then to forget stopping it. Discovery is expensive process that could  prevent device from entering power saving/efficient mode. 

I also agree that it is good to optimize the most common one "Find out whether there are available display(s), and if yes, show content on the display chosen by the user." I would like to make an addition to that description -  "if external device is not found then stop monitoring and forget that we ever wanted to use the second screen". 

Then there is a less common case that we still need to support - "Find out whether there are available display(s), and if yes, show content on the display chosen by the user." + keep monitoring for new devices because there could be better presentation destination appearing in the future. We should keep discovery process running in that case. Additionally, we  need to provide for developers an explicit mechanism for stopping discovery then.

After listing  those cases side by side I realized that I have already seen that approach implemented in Geolocation API [1]. There are three methods: to acquire position once, to start continuous watching and stopping the watch. We can follow the same pattern here:

```
partial interface NavigatorPresentation {
  Promise checkForAvailability(/*DOMString presentationUrl, params*/);
  Long watchForAvailability(/*DOMString presentationUrl, params*/);
  clearAvailabilityWatch( long watchId);
}
```
Method checkForAvailability runs discovery one time and stops after that. Last two methods would provide exactly the same functionality that is available in current design with callback attribute onavailablechange and corresponding event. New approach makes developers aware that browser starts doing extra work when monitoring in active. All that clarity comes with the price of adding new methods to proposed API. The other thing is that proposal mixes promises and callbacks together. I am sure, there might be a better way to design it.

[1] http://dev.w3.org/geo/api/spec-source.html#geolocation_interface

mwatson2:
Another common use-case is "Find out whether there are available
[...]

PR #89, opened 2015-05-19
Rethinking availability monitoring

avayvod:
The initial stab at the issue #81


mfoltzgoogle:
It looks like we have reached a consensus on #81, thank you for drafting a comprehensive PR to update the spec. 

I'll take a look at this change with some comments, can you rebase and post a new patch with responses?

avayvod:
I uploaded the rebased PR, will update it with the last comments.

avayvod:
I've updated the PR with the comments from the GH issue and also posted a unified text diff between the availability monitoring sections since these were moved and GitHub's diff is not very helpful.
The diff is here: http://avayvod.github.io/diff-before-after.txt
The updated spec is here: http://avayvod.github.io/availability.html

Overview.src.html
```diff
@@ -392,11 +392,26 @@ <h2 class="no-num no-toc">
...
+  // show or hide cast button depending on display availability
+  var handleAvailabilityChange = function(available) {
+    castBtn.style.display = available ? "inline" : "none";
   };
+  // Promise is resolved as soon as the presentation display availability is known.
+  navigator.presentation.getAvailability().then(function(availabilityObj) {
+    // availability.current may be kept up-to-date by the UA as long as the availability
+    // object is alive. It is advised for the web developers to discard of the object
+    // as soon as it's not needed (see the starting a new presentation example below).
+    availability = availabilityObj;
```
mounirlamouri: You don't need that. You could simply have ```availability``` as the argument in the then statement, like:
```js
navigator.presentation.getAvailability().then(function(availability) {
  // use |availability|.
});
```

Calling ```.onchange``` on the object should keep it alive as long as ```onchange``` isn't reset to ```null```.

index.html
```diff
@@ -886,7 +901,136 @@ <h4 id="event-handlers"><span class="secno">6.3.4 </span>
         </section>
       </section>
       <section>
-        <h3 id="interface-navigatorpresentation"><span class="secno">6.4 </span>
+        <h3 id="interface-availability"><span class="secno">6.4 </span>
+          Interface <a href="#availability"><code>Availability</code></a>
+        </h3>
+        <pre class="idl">interface <dfn id="availability">Availability</dfn> : EventTarget {
+  readonly attribute boolean current;
```
mounirlamouri: I don't know if ```current``` is the best name. I would consider ```value``` but it doesn't matter much I guess.

Overview.src.html
avayvod: I'm keeping the object in order to dereference it explicitly in setSession() below. Thus the page ensures the UA can stop monitoring availability. We can omit this from the example for simplicity.

index.html
avayvod: I wanted to avoid calling it ```available``` since it lives in the ```Availability``` interface :) however it make more sense if you consider using it like 

```javascript
availability.onchange = function() {
  handleAvailabilityChange(this.available); // vs. this.current or this.value
};
```

Otherwise I agree that ```value``` is a common name used along with ```onchange```.

avayvod:
Resolved the comments

Overview.src.html
```diff
@@ -392,11 +392,23 @@ <h2 class="no-num no-toc">
...
-    // show or hide cast button depending on display availability
-    castBtn.style.display = evt.available ? "inline" : "none";
+  // show or hide cast button depending on display availability
+  var handleAvailabilityChange = function(available) {
+    castBtn.style.display = available ? "inline" : "none";
   };
+  // Promise is resolved as soon as the presentation display availability is known.
+  navigator.presentation.getAvailability().then(function(availability) {
+    // availability.value may be kept up-to-date by the UA as long as the availability
+    // object is alive. It is advised for the web developers to discard of the object
```
mfoltzgoogle: It is advised for Web developers to discard the object...

Overview.src.html
```diff
@@ -392,11 +392,23 @@ <h2 class="no-num no-toc">
...
+    // as soon as it's not needed.
+    handleAvailabilityChange(availability.value);
+    availability.onchange = function() { handleAvailabilityChange(this.value); };
+  }.catch(function() {
+    // availability monitoring is not supported by the platform, discovery of second screens
```
mfoltzgoogle: nit: s/second screens/presentation displays/

Overview.src.html
```diff
@@ -392,11 +392,23 @@ <h2 class="no-num no-toc">
...
+    // will happen only after startSession() call, pretend the devices are available for
```
mfoltzgoogle: I would put "Pretend..." into a separate sentence

Overview.src.html
```diff
@@ -835,6 +847,147 @@ <h2 class="no-num no-toc">
...
+  attribute <span data-anolis-spec=
+"w3c-html">EventHandler</span> <span>onchange</span>;  
+};
+</pre>
+        <p>
+          The <code>change</code> event is fired during the execution of the
+          <span>monitoring presentation display availability</span> algorithm
+          when the <dfn>presentation display availability</dfn> changes. It is
+          fired at the <span><code>Availability</code></span> object, using the
+          <code>ChangeEvent</code> interface. The <code>Availability</code>
```
mfoltzgoogle: Where is ChangeEvent defined?

Overview.src.html
```diff
@@ -835,6 +847,147 @@ <h2 class="no-num no-toc">
...
+          object's <code>value</code> attribute is updated to the boolean value
+          that the algorithm determined.
+        </p>
+        <p>
+          In order to satisfy the <a href=
+          "https://github.com/w3c/presentation-api/blob/gh-pages/uc-req.md#req08-power-saving-friendly">
+          power saving non-functional requirement</a>, the user agent must keep
```
mfoltzgoogle: I would replace "must" with "should" as the requirement only applies to mobile UAs

Overview.src.html
```diff
@@ -835,6 +847,147 @@ <h2 class="no-num no-toc">
...
+          track of the number of <code>Availability</code> objects requested
+          through the <code>getAvailability()</code> method. The UA may also
+          keep track of whether the page holding the <code>Availability</code>
+          object is in the foreground. Using this information, implementation
+          specific discovery of <span title="presentation-display">presentation
+          displays</span> can be resumed or suspended, in order to save power.
+        </p>
+        <p>
+          The user agent must keep a <dfn>list of available presentation
```
mfoltzgoogle: Theoretically the UA does not need to maintain a list for this API.  It only needs to know if there is at least one display available.

Overview.src.html
```diff
@@ -835,6 +847,147 @@ <h2 class="no-num no-toc">
...
+              </ol>
+            </li>
+            <li>Set the <span>list of available presentation displays</span> to
+            the value of <a href="#newdisplays">newDisplays</a>.
+            </li>
+          </ol>
+          <p class="note">
+            The mechanism used to monitor <a href=
+            "#presentation-display">presention displays</a> availability is
+            left to the user agent. The user agent may choose to search for
```
mfoltzgoogle: But if the UA returns an Availability object it's obligated to keep the display status up to date.  Might rephrase as,

The UA may choose to monitor the availability of presentation displays even when there are no Availability objects.

Overview.src.html
```diff
@@ -835,6 +847,147 @@ <h2 class="no-num no-toc">
...
+            displays at any time, not just when there's an alive
+            <code>Availability</code> object.
+          </p>
+          <p>
+            When the user agent is to <dfn>cancel monitoring the list of
```
mfoltzgoogle: What triggers the running of these steps?

Overview.src.html
```diff
@@ -1181,7 +1333,67 @@ <h2 class="no-num no-toc">
         </section>
         <section>
           <h4>
-            The <code>onavailablechange</code> EventHandler
+            Getting the <span>presentation screens</span> availability
```
mfoltzgoogle: <span>presentation display<span>s

mfoltzgoogle:
Thanks! Some minor edits suggested above.  If you prefer, rebase and I can merge, then make a copy editing pass.

Overview.src.html
avayvod: Done.

Overview.src.html
avayvod: Actually nothing. I just moved this chunk with the rest of the availability algorithms to the Availability object. Now before the change this would be triggered when there's no onavailablechange event listeners but then the algorithm doesn't make sense since it only fires the events. Removed.

Overview.src.html
avayvod: Done.

Overview.src.html
avayvod: No, but the current list is used for example in the algorithm of change detection (comparing the currently known list with the new list found during the next iteration of continuous monitoring). We can rephrase that to make the spec more flexible but I'd rather do it in a separate low-priority pull request tracked by another issue.

Overview.src.html
avayvod: Done.

Overview.src.html
avayvod: Done.

Overview.src.html
avayvod: Done.

Overview.src.html
avayvod: http://www.w3.org/TR/html5/webappapis.html#idl-definitions

Overview.src.html
avayvod: Done

Overview.src.html
avayvod: Done

Overview.src.html
avayvod: Done.

avayvod:
Rebased and updated the PR. Replied to the comments. PTAL.

Overview.src.html
mfoltzgoogle: Ideally we would add this document as a reference here.  This can be done in a subsequent PR.

Overview.src.html
```diff
@@ -830,6 +842,126 @@ <h2 class="no-num no-toc">
...
+          track of the number of <code>Availability</code> objects requested
+          through the <code>getAvailability()</code> method. The UA may also
+          keep track of whether the page holding the <code>Availability</code>
+          object is in the foreground. Using this information, implementation
+          specific discovery of <span title="presentation-display">presentation
+          displays</span> can be resumed or suspended, in order to save power.
+        </p>
+        <p>
+          The user agent must keep a <dfn>list of available presentation
+          displays</dfn>. According to the number of alive
```
mfoltzgoogle: We should define what it means for an Availability object to be "alive" - I assume "not eligible for garbage collection."   I wonder if there is a standard definition for this?

Overview.src.html
```diff
@@ -1175,7 +1306,67 @@ <h2 class="no-num no-toc">
...
+            "promise-objects">Promise</span>.
+            </li>
+            <li>Return <em>P</em>.
+            </li>
+            <li>If the user agent doesn't support continuous monitoring of the
+            presentation display availability, then:
+              <ol>
+                <li>
+                  <span data-anolis-spec="promguide" title=
+                  "resolve-reject">Reject</span> <em>P</em>.
```
mfoltzgoogle: Can reject with a standard DOMError like NotSupportedError:

http://www.w3.org/TR/domcore/#error-names-0

mfoltzgoogle:
I made a few more inline comments but this is a good solid draft.  The remaining items can be addressed in a later PR.

Issue #76, opened 2015-04-10
URL schemes supported in presentation API


louaybassbouss:
Presentation API addresses currently presentation of hosted web content (e.g. accessible via `http://` or `https://`). According to the spec, presentation URLs follows the [URL Standard](https://url.spec.whatwg.org/). Questions is: are there other relevant URL schemes for Presentation API? for example `data:` scheme can be used in the following way:

``` javascript
var presentationUrl = 'data:text/html,<!DOCTYPE html><html><body><p>Test</p><script>/* inline JavaScript that may use navigator.presentatoion.session */</script></body></html>';
var session = navigator.presentatoion.startSession(presentationUrl);
``` 

Also the `file://` protocol could be relevant for non-hosted web content for example for Packaged Web Application like [W3C Widgets](http://www.w3.org/TR/widgets/).

All relevant schemes need to be evaluated regarding limitations, security implications, performance etc. for usage in Presentation API. Also use cases to derive requirements are needed.

louaybassbouss:
@anssiko comment on this:

>That is a good question. Whatever we decide, we should specify the expected behaviour in order to allow the implementations to interoperate after we've figured out whether the API should work with data URIs.

> I think there may be a disagreement among implementers whether data URIs are same-origin depending on where they're used (Worker, iframe, img, open()...). I think the closest existing API to startSession() where data URIs can be used is open().

> I think the primary motivation for using data URIs has been performance. HTTP/1.x has known limitations in loading multiple resources concurrently, so inlining all the resources has improved the performance. Another use case I can think of is to be able to provide a single file that contains all the resources without the need to package multiple files together.

> Louay, All - I think you should investigate this further and come up with a recommendation on how to proceed i.e. whether data URIs should be allowed or not, and how to spec that

mfoltzgoogle:
Another open question is whether vendor- or OS-specific schemes are allowed.  For example, a Chrome extension or app may want to present content which would be hosted by that extension/app (using the scheme `chrome-extension://`).  Or a browser on a mobile device may want to present content that is available to a native application on the device (via some OS-specific scheme).

These scenarios may just be user agent specific features that are a superset of the specification, and not part of the specification itself.

anssiko:
PROPOSED RESOLUTION: do not restrict extensions that might want to make use of other URL schemes than http(s), however those alternative schemes are out of scope for v1

Relevant discussion at Berlin F2F: http://www.w3.org/2015/05/20-webscreens-minutes.html#item08

[...]

PR #96, opened 2015-05-27
Add NOTE clarifying presentation URL scheme support

mfoltzgoogle:
Updates spec for proposed resolution of Issue #76.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -1176,10 +1176,10 @@ interface <dfn>NavigatorPresentation</dfn> : EventTarget {
           </p>
           <p class="note">
             The <code>presentationUrl</code> should name a resource accessible
-            to the local or a remote user agent. User agents must accept a
-            <code>presentationUrl</code> using the <code>http</code> or
-            <code>https</code> schemes; behavior for other schemes is not
-            defined by this specification.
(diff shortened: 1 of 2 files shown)


tidoust:
The sentence "User agents must accept a presentationUrl using the http or https schemes" sounds very normative to me, which looks strange in a note that is informative-only by definition.

I suggest to rewrite to:
```html
<p class="note">The <code>presentationUrl</code> should name a resource accessible
to the local or a remote user agent. This specification defines behavior for
<code>presentationUrl</code> using the <code>http</code> or <code>https</code>
schemes; behavior for other schemes is not defined by this specification.</p>
```

Note that user agents will de facto be required to support http and https schemes because algorithms will mandate behavior in such cases. As such, there should be no need to add an extra normative statement such as "user agents MUST support presentationUrl using the http or https schemes".

The "should" in "The presentationUrl should name a resource accessible" could also be read as normative but it does not apply to user agents, so less of a problem.

louaybassbouss:
@mfoltzgoogle the suggestion from @tidoust LGTM

mfoltzgoogle:
Thanks @tidoust and @louaybassbouss.  I've updated the PR with the suggested text and will now merge.

PR #95, opened 2015-05-27
Update requirements for multiple controllers and presentations

mfoltzgoogle:
Updates requirements to allow many-to-one and one-to-many relationships between controllers and presentations.

Updates terminology to use "controller" consistently for the opening/joining browsing context.

PR #101, opened 2015-06-01
Fix bugs in availablechange event

mfoltzgoogle:
Addresses Issue #95

* Make sure we designate a target for the `availablechange` event
* Fix confusion between event and handler


mfoltzgoogle:
This seems like a straightforward change, so merging and closing.

Issue #45, opened 2015-01-30
Security and privacy considerations


anssiko:
We should complete the Security and privacy considerations section to the Presentation API once we're better aware of the threats.

The spec has been evaluated against [Self-Review Questionnaire: Security and Privacy][2], with results documented in this issue. Furthermore, the Privacy Interest Group (PING) has produced documentation that may be helpful, namely [Fingerprinting Guidance for Web Specification Authors][3] and [Privacy Considerations for Web Protocols][4]. The [Web Security IG review][5] and [PING review][6] of the specification have been initiated. This sections should be crafted based on feedback from the respective reviews.

[1]: http://www.w3.org/2014/secondscreen/charter.html#scope
[2]: https://w3ctag.github.io/security-questionnaire/
[3]: https://w3c.github.io/fingerprinting-guidance/
[4]: https://w3c.github.io/privacy-considerations/
[5]: https://lists.w3.org/Archives/Public/public-web-security/2015Jul/0000.html
[6]: https://lists.w3.org/Archives/Public/public-privacy/2015JulSep/0003.html

anssiko:
https://lists.w3.org/Archives/Public/public-secondscreen/2015Jan/0037.html

anssiko:
We should evaluate the Presentation API using the Strawman Self-Review Questionnaire: Security and Privacy [1] kindly authored by @mikewest. It is preferred to do this sooner rather than later, and for example, present and discuss the results of the evaluation at F2F.

All - please let us know if you'd like to help evaluate the spec using the questionnaire [1].

[1] https://mikewest.github.io/spec-questionnaire/security-privacy/

tidoust:
I'm assigning the issue to myself as no one seems to have jumped on it. I do not claim to have any particular expertise on security and privacy issues but I'll try to follow the existing guidelines and questionnaire, and to use common sense. I'll share evaluation results before the F2F.

anssiko:
Depending on the results of the security and privacy evaluation, the Privileged Contexts [1] spec may become relevant:

>[The Privileged Contexts] specification provides guidelines for user agent implementors and spec authors for implementing features whose properties dictate that they be exposed to the web only within a trustworthy environment.

It seems the Presentation API ticks some of the boxes in *Should [insert feature here] require a privileged context?* [2]:
* The feature provides access to or information about other devices a user has access to.
* The feature introduces some functionality for which user permission will be required.

[1] https://w3c.github.io/webappsec/specs/powerfulfeatures/
[...]

PR #104, opened 2015-06-02
Issue #45: Security and Privacy Considerations

mfoltzgoogle:
This PR drafts the Security and Privacy Considerations following discussion in #45.  @tidoust and others, please review.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -1436,24 +1436,20 @@ dictionary <dfn>DefaultSessionStartEventInit</dfn> : EventInit {
       </h3>
       <p>
         A <span>presentation session</span> is allowed to be accessed across
-        origins; the URL that started the presentation and the presentation ID
-        are the only information needed to join a session from any origin in
-        that user agent. In other words, a presentation is not tied to a
-        particular opening origin.
+        origins; the presentation URL and presentation ID used to create the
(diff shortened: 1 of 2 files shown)


tidoust:
Please find a few comments below. The rest looks good. I would merge the pull request in any case, the text provides a very good basis for that section!

### Cross-origin access
* "the URL that started the presentation": couldn't this mean the URL of the opening context? Change to "the URL of the presentation session", perhaps?
* I would drop the paragraph starting from "We could further restrict...". We're not going to do that in practice, so I don't see what it brings. Perhaps rephrase the whole paragraph as: "This design allows controlling contexts from different domains to connect to a shared presentation resource. The security of the presentation ID prevents arbitrary pages from connecting to an existing presentation.".
* I would drop "the charter envisions", or use "the group envisions" instead (most readers won't know what a charter is and that's not a useful concept in a spec)

### Temporary identifiers and browser state
I would drop "Again, one possible solution would be to restrict the API to secure contexts" since my understanding is that we would prefer not to do that, and replace with an open issue such as:
```html
<p class="open-issue">
  Should we restrict the API to some extent in non secure contexts?
</p>
```

mfoltzgoogle:
All good suggestions @tidoust.  I have incorporated them into the updated PR and will now merge.

Issue #39, opened 2015-01-12
Resumption of multiple sessions


tidoust:
## Description
If a page created two different sessions for the same URL and presentation ID on two different screens, how can it reconnect to both sessions?

The current algorithm for reconnecting sessions returns the session that was most recently connected to.

## Proposal
Should `joinSession` rather return an array of `PresentationSession` objects?
Should the page rather use different presentation IDs?

## References
See initial discussion in https://lists.w3.org/Archives/Public/public-secondscreen/2015Jan/0012.html

mfoltzgoogle:
A related issue is when the page calls `startSession(url, id)` when there is an existing presentation with the same `url` and `id`.  The [current algorithm](http://w3c.github.io/presentation-api/#starting-a-presentation-session) indicates that a new presentation session is always started, but the fate of the existing session is not specified.  The user agent could:

1. Keep the existing session, which would allow the scenario posted originally in this issue to occur.  However, `joinSession` and other algorithms would need to be modified to handle the `(url, id)` ambiguity.
2. Close the existing session only if a different screen was chosen.  Otherwise act as if `joinSession` was called and reconnect to the existing presentation.
3. Always close the existing session (even if the same screen was chosen)

2 seems to match the intent of the user in most cases (i.e. redirecting a presentation to a different screen), as opposed to multicasting an identical presentation to multiple screens.

tidoust:
See related discussion at F2F:
http://www.w3.org/2015/05/20-webscreens-minutes.html#item06

PROPOSED RESOLUTION: drop id from startSession() without strong use case, the UA generates the id

mfoltzgoogle:
Closing this issue as PR #109 implements the resolution from the F2F, and there hasn't been further comment.

PR #109, opened 2015-06-09
Issue #39: Drop presentationId from startSession

mfoltzgoogle:
This PR implements the action from #39: drop id from startSession() without strong use case, the UA generates the id


mfoltzgoogle:
Note: I will rebase with #108 before merging.

avayvod:
Could you also remove the id from the default presentation link element in the example?

mfoltzgoogle:
@avayvod: Done.

PR #110, opened 2015-06-10
Use HTTPS where possible

mathiasbynens:
This avoids mixed content issues on https://w3c.github.io/presentation-api/.


anssiko:
Thanks, good catch. Fixed via https://github.com/w3c/presentation-api/pull/111 (updated the Anolis source too).

PR #117, opened 2015-06-17
Second version of Issue #9: How to filter available screens according to the content being presented

mfoltzgoogle:
This PR is an amendment to #114 that makes the URL argument mandatory to `getAvailability`, which simplifies the algorithm description.  It also updates the algorithm for `startSession` to take per-URL availability into account.


Overview.src.html
```diff
@@ -850,90 +852,128 @@ <h2 class="no-num no-toc">
...
-          title="resolve-reject">rejected</span> and the algorithm to
-          <span>monitor the list of available presentation displays</span> will
-          only run as part of the <span title="startSession">session
-          start</span> algorithm.
-        </p>
         <section>
-          <h5>
-            Monitoring the list of available presentation displays
-          </h5>
+          <h4>
```
avayvod: Nit: Don't you want to have a section for each header to make it accessible from TOC?

Overview.src.html
```diff
@@ -850,90 +852,128 @@ <h2 class="no-num no-toc">
...
           <p>
-            While there are alive <code><span>Availability</span></code>
-            objects, the user agent should continuously keep track of the
-            <span>list of available presentation displays</span> and repeat the
-            following steps:
+            The user agent must keep a <dfn>list of available presentation
+            displays</dfn>. This current list of <span>presentation
+            displays</span> that may be used for starting new presentations,
+            and is populated based on an implementation specific discovery
+            mechanism. It is set to most recent result of the algorithm to
```
avayvod: Nit: s/most/the most

Overview.src.html
```diff
@@ -850,90 +852,128 @@ <h2 class="no-num no-toc">
...
+            "https://github.com/w3c/presentation-api/blob/gh-pages/uc-req.md#req08-power-saving-friendly">
+            power saving non-functional requirement</a>. To further save power,
+            the UA may also keep track of whether the page holding an
+            <code>Availability</code> object is in the foreground. Using this
+            information, implementation specific discovery of <span title=
+            "presentation-display">presentation displays</span> can be resumed
+            or suspended.
+          </p>
+          <p>
+            Some <span>presentation display</span>s may only be able to display
```
avayvod: Not: don't break display and s to be consistent with a similar span before?

Overview.src.html
```diff
@@ -850,90 +852,128 @@ <h2 class="no-num no-toc">
...
+            <dfn>display availability URL</dfn> if the user agent can
+            reasonably guarantee that the presentation of the URL on that
+            display will succeed.
+          </p>
+          <h4>
+            Monitor the list of available presentation displays
+          </h4>
+          <p>
+            If <span>set of availability objects</span> is non-empty, or there
+            is a pending call to <span title="startSession">start a
```
avayvod: Not: s/call/request?
Request to start a session reads better imho than call to start a session. I would use call with a method name, e.g. call to startSession()

Overview.src.html
```diff
@@ -850,90 +852,128 @@ <h2 class="no-num no-toc">
...
+            available presentation displays</dfn> by running the following
+            steps.
           </p>
           <ol>
             <li>
               <span data-anolis-spec="w3c-html">Queue a task</span> to retrieve
-              the list of currently available <a href=
-              "#presentation-display">presentation displays</a> and let <a id=
-              "newdisplays"><em>newDisplays</em></a> be this list.
+              available presentation displays (using a implementation specific
```
avayvod: Not: s/a/an

Overview.src.html
```diff
@@ -941,17 +981,27 @@ <h2 class="no-num no-toc">
...
+            is eligible for garbage collection), the user agent should run the
+            following steps:
+          </p>
+          <ol>
+            <li>Find and remove any entry <em>(A, availabilityUrl, P)</em> in
+            the <span>set of availability objects</span> for the newly deceased
+            <em>A</em>.
+            </li>
+            <li>If the <span>set of availability objects</span> is now empty
+            and there is no pending call to <span title="startSession">start a
```
avayvod: Nit: ditto regarding call and start a session

Overview.src.html
```diff
@@ -1013,9 +1063,26 @@ <h2 class="no-num no-toc">
             </li>
             <li>Return <em>P</em>.
             </li>
-            <li>If the user agent does not <span>monitor the list of available
-            presentation displays</span>, run the following steps:
+            <li>Run the following steps:
               <ol>
+                <li>If is an entry <em>(A, presentationUrl, P)</em> in the
```
avayvod: Nit: s/If is/If there is

Overview.src.html
```diff
@@ -1013,9 +1063,26 @@ <h2 class="no-num no-toc">
...
+                <span>set of availability objects</span>, where
+                <em>presentationUrl</em> is equal to
```
avayvod: Not sure how it looks in the generated html but I think it would be hard to distinguish between presentationUrl and presentationUrl. Maybe call the first one availabilityUrl?

avayvod:
Lgtm with minor nits. Could you host the generated for somewhere to preview the results of the change, please.

mfoltzgoogle:
Thanks for the detailed review.

The updated version is hosted here: https://rawgit.com/w3c/presentation-api/filter-screens-issue-9-noundef/index.html

Two specific responses:

> Not: don't break display and s to be consistent with a similar span before?

I converted similar usages to not break the plural, but it means Anolis won't automatically convert it the `<span>` to a link to the definition.  Personally I think this reads better though.  If we really want to we can mark up the `<span>` tags to fix up the references.

> Not sure how it looks in the generated html but I think it would be hard to distinguish between presentationUrl and presentationUrl. Maybe call the first one availabilityUrl?

I simplified the start session algorithm to not refer to Availability objects. They aren't really needed to implement startSession, the browser just needs to ensure there is at least one compatible display.

mfoltzgoogle:
I'd like to merge this and take another pass later, to perhaps improve the wording of a few sections, once I've had a chance to address other pending issues.  Issue #9 will remain open while the privacy considerations are under discussion and additional feedback can happen there.

Issue #91, opened 2015-05-25
Separate interface for controlling page and presenting page


schien:
Currently we mix the API for controlling page and presenting page in `NavigatorPresentation` and developer will use `navigator.presentation` as the entry point.

Since we are providing more and more advance feature in this API (see #19), it's cumbersome to provide statement on the usage of each attribute/function.

Therefore I propose to move all presenting page APIs into a different interface, called `PresentingContext`. The example definition will looks like:

```
interface PresentingContext : EventTarget {
  Promise<PresentationSession> getSession();
  Sequence<PresentationSession> getSessions();
  attribute EventHandler onsessionavailable;
};

partial interface Navigator {
  readonly attibute PresentingContext presentingContext; // null if not presenting page
};
```

anssiko:
@schien Thanks for the concrete proposal.

All - @tidoust has a nice explanation in the [Conformance classes definition][1] issue why this separation makes very much sense. Please read the issue, and let the group know if you have any concerns with separating the interfaces to reflect the two conformance classes of products as follows (naming subject to change): *opening user agent* (or *controlling user agent* to align with #95) and *presenting user agent*. This issue proposes the first concrete change to implement this separation.

@mfoltzgoogle How would you like to proceed with baking this into the spec? I suggest you coordinate with @schien who volunteered to help.

[1]: https://github.com/w3c/presentation-api/issues/93

mfoltzgoogle:
I think I would like to define the conformance classes for the controlling/opening browsing context and the presenting browsing context as in #93 and re-outline the document as best as I can along those lines.

Then it will be easier to introduce the new interfaces proposed by @schien since they will have a logical place in the new document.

avayvod:
A naming idea:
navigator.present - for the controller (sender) page
navigator.presentation - for the presentation (receiver) page

So we leave what we have for the presentation where it is and move the sender API to the shorter interface.

anssiko:
It seems no concerns raised with the general direction proposed by @schien. It also appears we've not yet settled on the interface naming. We should first update the Conformance section using the [@tidoust's proposal][1] as a starting point and bake in the changes to the normative parts when consensus emerges on naming.

All - please let us know your preference and/or suggestions with respect to naming.

@mfoltzgoogle Feel free to update the Conformance section with a note stating that the rest of the spec have not yet been reworked to match the conformance classes split.

[1]: https://github.com/w3c/presentation-api/issues/93

mounirlamouri:
[...]

Issue #19, opened 2014-09-03
Specify behavior when multiple controlling pages are connected to the session


mfoltzgoogle:
An implication of reconnection is that multiple pages may connect to the same presentation.  The behavior of the API should be defined for this scenario.  Specifically,

* Firing of `onstatechange` = `"connected"`/`"disconnected"` on both sides
* Semantics of `close()` on both sides

mfoltzgoogle:
Adding to this issue from #36 discussion of how messaging should be done between the presentation and multiple connected sessions.

In some scenarios the presented page may have active connections to multiple PresentationSessions on other pages. In a gaming scenario, each connected session would represent a game player. Some information (like dealing a card into the hand) should be sent to one player, while other information (playing a card face up) should be sent to all players.

We need to specify:
1. How the presented page may communicate to a specific connected page
2. How the presented page may broadcast - communicate to all specific connected pages

Note [this thread](http://lists.w3.org/Archives/Public/public-secondscreen/2014Nov/0047.html) discussing some concrete spec changes to facilitate this case.

jakearchibald:
The `SharedWorker` model of a `connect` event that provides access to the connecting client may help here.

anssiko:
Agreed with @jakearchibald on that. Would you @mfoltzgoogle with help from @drott like to investigate this further?

tidoust:
PROPOSED RESOLUTION: starting point for #19 is Promise<PresentationSession> NP.getSession(), PresentationSession[] NP.getSessions(), NP.onsessionavailable event when set of sessions changes (with possible naming changes)

See related discussion at F2F in Berlin:
http://www.w3.org/2015/05/19-webscreens-minutes.html#item12

anssiko:
ACTION: @mfoltzgoogle to look at renaming "sessions" for controlling and presenting sides. [recorded in http://www.w3.org/2015/05/19-webscreens-minutes.html#action12] 

ACTION: @mfoltzgoogle to fix spec to refer to updated Presentation idiom [recorded in http://www.w3.org/2015/05/19-webscreens-minutes.html#action13]

anssiko:
Also, should define the mechanism by which the page can create a new session that can access existing presentation.

avayvod:
How about the following spec change for the multiple connections? It's based on the assumption that only one controller can connect to the presentation page at a time. If the page is slow handling each controller, the UA can simply queue them and resolve the next call to ```getNextController``` when needed.

```idl
[...]

PR #118, opened 2015-06-17
Initial description of session creation and monitoring in presenting …

obeletski:
Initial description of session creation and monitoring in presenting browsing context. Methods `getSession()` and `getSessions()` added to `NavigatorPresentation` interface as agreed at F2F http://www.w3.org/2015/05/19-webscreens-minutes.html#action12. 
Interface still might change since discussion is ongoing in https://github.com/w3c/presentation-api/issues/19. This PR is a starting point for future modifications coming from https://github.com/w3c/presentation-api/issues/91 as well.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -965,7 +965,8 @@ partial interface <span data-anolis-spec="w3c-html">Navigator</span> {
         </p>
         <pre class="idl">
 interface <dfn>NavigatorPresentation</dfn> : EventTarget {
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
@obeletski  thank you for drafting an initial PR to specify this algorithm.  I will pull it into a local branch, resolve merge conflicts, and make some small copy editing improvements. 

I will also add a reference to Issue #19, as I think there will be ongoing discussion about the shape of this API.

PR #119, opened 2015-06-17
fix send algorithm

jcdufourd:
I believe the send algorithm should mention ArrayBufferView in binary mode.

### DIFF
--- a/index.html
+++ b/index.html
@@ -726,7 +726,7 @@ interface <dfn id="presentationsession">PresentationSession</dfn> : EventTarget 
             </li>
             <li>Let <a href="#presentation-message-type">presentation message type</a> <em>messageType</em>
             be <code>binary</code> if <code>data</code> is one of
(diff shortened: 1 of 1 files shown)


jcdufourd:
If the underlying object is the same for ArrayBuffer and ArrayBufferView, then my PR is not relevant. Otherwise, there is a need for a larger fix, inclusion of a binarytype arraybufferview and modification of reception algorithm too.

mfoltzgoogle:
Yes, thank you for catching this, LGTM.  

However, the patch needs to be made to Overview.src.html, then index.html is generated from that and they are committed together.

This wiki describes the process:  https://www.w3.org/wiki/Second_Screen/Work_Mode#How_to_Contribute_via_Pull_Requests

Note it sounds like we will be moving to a new document system where this will be easier (i.e. no makefiles).

I will make this same fix in a different PR.

Issue #68, opened 2015-03-25
Review the use cases and requirements


anssiko:
The Presentation API attempts to address the use cases and requirements outlined in the specification at: http://w3c.github.io/presentation-api/#use-cases-and-requirements-0

We should evaluate the UCs and Reqs we laid out initially, and based on group discussion and consensus see if changes are needed. For example, people working on the implementations may have discovered new requirements that we did not anticipate initially that would be in scope for the group. I propose we discuss this at F2F time permitting, and to make for a productive session, more complete proposals for new use cases and requirements should be provided before the meeting.

hongkicha:
Anssi, we currently have four UCs so why don't we add another category? **Digital Signage(DS)**. <br>DS is a screen media showing contents such as advertisements and notifications. <br>DS is one of the closest media we bump into everyday.<br> I strongly believe Presentation API should support some functionalities DS requires. <br>  In this sense, I would like to...
1. update our document by adding a short description on DS.
2. prepare a not too long presentation on DS at our upcoming F2F meeting.
Your thoughts?

anssiko:
@hongkicha Thanks for volunteering to look at how the Presentation API might be used in digital signages.

To start, I suggest you prepare such a document outside of this repository and provide a link in here. Please note that digital signage specific features are out of scope for this Working Group. Such features are better discussed and specified in the Second Screen Presentation Community Group [1] that is set out to incubate experimental ideas, or in the Web-based Signage Business Group [2], that has produced use cases [3].

Requirements that are deemed generic may motivate changes to the Presentation API. Identifying such generic requirements would be something I'd like to see discussed at F2F in the UCs and Reqs slot to keep the group focused on its Scope [4].

[1] https://www.w3.org/community/webscreens/
[2] https://www.w3.org/community/websignage/
[3] http://www.w3.org/community/websignage/wiki/Web-based_Signage_Use_cases_and_Requirements
[4] http://www.w3.org/2014/secondscreen/charter.html#scope

hongkicha:
@anssiko Thanks Anssi for clarification. Now I got to know more about how our WG actually works. I thought it was fine to introduce DS directly here at WG. I'll get back to this issue soon. Thank you again!

anssiko:
ACTION: Louay to initiate the Use cases and Requirements document (probably on a Wiki page) [recorded in http://www.w3.org/2015/05/19-webscreens-minutes.html#action01]

anssiko:
[...]

PR #126, opened 2015-06-26
Issue #68: Use cases and Requirements document updated

louaybassbouss:
Issue #68: Use cases and Requirements document updated as following:
* Update "UC02: Video and image Sharing" to derive requirement "REQ03: Resuming presentation".
* Add missing requirement "REQ04: Joining presentation". This is derived from the multiplayer game use case
* Update the requirements derived from each use cases (each requirement is addressed in at least one use case).
* Start numbering for non-functional requirements with NF-REQ01. This allows us the add new functional and non-functional requirements without mixing the numbers.

TODO: in Overview.src.html change "https://github.com/w3c/presentation-api/blob/gh-pages/uc-req.md#req08-power-saving-friendly"  into "https://github.com/w3c/presentation-api/blob/gh-pages/uc-req.md#nf-req01"

### DIFF
--- a/uc-req.md
+++ b/uc-req.md
@@ -10,7 +10,7 @@ This section collects relevant use cases for the [Presentation API][pres-api].
 
 > Each use case is defined in its own subsection. To define a new use case just
 > copy/paste an existing one and update the use case identifier (UCxx), title,
-> dependencies, description and requirements.
+> description and requirements.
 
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
LGTM.  I will send another PR to update the power saving requirement in the specification document.

Issue #121, opened 2015-06-18
Section " Messaging between presentation sessions" should be updated


obeletski:
Section [1] should be updated or dropped because
- it repeats what is already stated in [2]
- Messaging is happening between the contexts, not session [3]

1. http://w3c.github.io/presentation-api/#messaging-between-presentation-sessions
2. http://w3c.github.io/presentation-api/#sending-a-message-through-presentationsession
3. http://w3c.github.io/presentation-api/#common-idioms

mfoltzgoogle:
@obeletski, it looks like PR #124 edited these sections to address the two issues that you mention above.  Can this issue be closed now?

PR #129, opened 2015-06-29
Update to messaging section

obeletski:
Update to messaging section to reflect the fact that messaging is happening between presenting and controlling contexts, https://github.com/w3c/presentation-api/issues/121.

### DIFF
--- a/Overview.src.html
+++ b/Overview.src.html
@@ -1773,11 +1773,13 @@ dictionary <dfn>DefaultSessionStartEventInit</dfn> : EventInit {
         presentations in "incognito" (private browsing context) mode.
       </p>
       <h3>
-        Messaging between presentation sessions
+        Messaging between contexts
       </h3>
       <p>
-        This spec will not mandate communication protocols, but it should set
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
This would imply that messaging is happening between Documents or tabs - see the definition of browsing context [1].

However the spec should be clear that a message passed to one PresentationSession must be sent to a corresponding PresentationSession, and there could be multiple per document.

[1] 

Maybe this could be written as, 

```
This spec will not mandate communication protocols between the
<span>controlling browsing context</span> and the <span>presenting
browsing context</span>, but it should set some guarantees of message
confidentiality and authenticity between <span title="presentation session">
presentation sessions</span>.
```

obeletski:
Looks good, I would add one more "corresponding" word in that text.
```
This spec will not mandate communication protocols between the
<span>controlling browsing context</span> and the <span>presenting
browsing context</span>, but it should set some guarantees of message
confidentiality and authenticity between corresponding 
<span title="presentation session">presentation sessions</span>.
```

PR #130, opened 2015-06-29
Update to resuming requirement

obeletski:
Update to reflect the fact that we resume connection to the presentation and not the presentation itself, 
https://github.com/w3c/presentation-api/issues/68

### DIFF
--- a/uc-req.md
+++ b/uc-req.md
@@ -128,12 +128,12 @@ at the request of the UA. A single controller may be able to send content to
 multiple displays at once, and a single display may be able to host multiple
 presentations at once.
 
-### <a name="req03"></a>REQ03: Resuming presentation
+### <a name="req03"></a>REQ03: Resuming connection to presentation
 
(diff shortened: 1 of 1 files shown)


uc-req.md
```diff
@@ -128,12 +128,12 @@ at the request of the UA. A single controller may be able to send content to
...
 presentations at once.
 
-### <a name="req03"></a>REQ03: Resuming presentation
+### <a name="req03"></a>REQ03: Resuming connection to presentation
 
 If a controller becomes disconnected from its presentation, the UA must provide
 a way for it to resume its connection to the presentation. It must also provide
 a way for a new browsing context on the same UA to request connection to one or
-more presentations (thus becoming their controller).
+more presenting browsing contexts (thus becoming their controlling browsing context).
```
mfoltzgoogle: We define the terms "presentation" and "controller" in REQ02, let's continue using those in the rest of the text.

Otherwise LGTM

PR #133, opened 2015-07-03
Rename Availability interface to PresentationAvailability.

mounirlamouri:
The interface name is exposed in the global object so having it prefixed
with Presentation is probably safer. Also, it prevents the Presentation
API from taking ownership of the generic 'Availability' interface name.


avayvod:
lgtm
@anssiko @tidoust would one of you be able to merge this? Mark is OOO until Monday AFAIK.

tidoust:
Good point, thanks @mounirlamouri !

PR #134, opened 2015-07-06
Specify the DOMException name when rejecting a getAvailability() call.

mounirlamouri:
This is also including a typo fix.


mfoltzgoogle:
lgtm, thanks!

Issue #143, opened 2015-07-07
Update PresentationAvailability.value and fire the change event in the same task


foolip:
http://w3c.github.io/presentation-api/#interface-presentationavailability

> Queue a task to fire an event named change at availability object.

Changing the observable attribute and firing the event in separate tasks allows for the new value to be seen before the event is fired. This kind of mismatch can be tricky, for example it means that there's no guarantee that looking at the `value` attribute in the event handler there's no guarantee that one will see alternating states, for example if a device is available for just a millisecond it ought to be possible to observe `value` to be false both before and in both of the events.

Maybe just fire the event after resolving the promise?

mfoltzgoogle:
I believe that the way that the algorithm is written, each state transition for A.value should result in a change event fired on A with the corresponding value; if A.value transitions to "false" then "true" then two events should be fired, the first with A.value == false and the second with A.value == true.

It does seem cleaner to fire the event before resolving the promise, although the caller may or may not see the first change event depending on when the event task completes vs. adding the initial event handler.

In a coming PR, I made the change above, and also consolidated the event firing task into the main algorithm to make it easier to follow.

Issue #142, opened 2015-07-07
Missing "attribute" in "readonly DOMString? id;"


foolip:
https://w3c.github.io/presentation-api/#interface-presentationsession

foolip:
The keyword is also missing in "EventHandler onmessage;"

Issue #141, opened 2015-07-07
Replace PresentationSession[] with sequence<PresentationSession>


foolip:
http://heycam.github.io/webidl/#idl-sequence

The `T[]` syntax was removed from Web IDL today:
https://github.com/heycam/webidl/commit/416ebfb2c538363c1f2a763d1b2921f386b09e2e

Issue #140, opened 2015-07-07
Section DefaultSessionStart should be DefaultSessionStartEvent


foolip:
http://w3c.github.io/presentation-api/#interface-defaultsessionstart

Compare to https://w3c.github.io/uievents/#interface-MouseEvent

PR #144, opened 2015-07-07
Addresses issues raised by @foolip

mfoltzgoogle:
Addresses several issues raised by @foolip:

- #137: Rename NavigatorPresentation to just Presentation
- #139: Use [SameObject] for Navigator.presentation
- #140: Section DefaultSessionStart should be DefaultSessionStartEvent
- #141: Replace PresentationSession[] with sequence<PresentationSession>
- #142: Missing "attribute" in "readonly DOMString? id;"
- #143: Update PresentationAvailability.value and fire the change event in the same task


Overview.src.html
```diff
@@ -988,16 +976,23 @@ <h2 class="no-num no-toc">
...
-                availability change</span> with <em>A</em>.
+                <em>newAvailability</em>, then <span data-anolis-spec=
+                "w3c-html">queue a task</span> to run the following steps:
+                  <ol>
+                    <li>Set <em>A</em>'s <code>value</code> property to
+                    <em>newAvailability</em>.
+                    </li>
+                    <li>
+                      <span data-anolis-spec="w3c-html" title=
+                      "concept-event-fire">Fire an event</span> named
```
foolip: Perhaps this could be stated as either resolving the promise or firing the event, to make it very clear that firing the event wouldn't have any effect, so that we can avoid having the discussion of the ordering of promise and event.

foolip:
Changes LGTM, commented on two not-so-obvious things.

mfoltzgoogle:
It turns out Anolis gets confused by the definition of the term "Presentation" in the Common Idioms section [1] and won't link references to the `Presentation` interface definition without an explicit `<dfn>` tag marking it up.

We are switching authoring tools soon #116 and hopefully this can be cleaned up as part of the transition.  Will leave a TODO in the document.

[1] http://w3c.github.io/presentation-api/#common-idioms

mfoltzgoogle:
Made the other recommended change and will land this.  Thanks for the feedback @foolip!

foolip:
Thanks!

Issue #86, opened 2015-05-14
Define algorithm for browser initiated presentation using default presentation URL


mfoltzgoogle:
Pull request #77 added definitions to the spec for the default presentation URL and ID to allow the UA to initiate presentation on behalf of the page.  This issue tracks an update to the algorithms for 6.4, "Interface NavigatorPresentation" to define the side effects on the UA state when this happens.

mwatson2:
What causes the UA to initiate a session with the default presentation URL ?

mwatson2:
OK, I caught up ;-) The default presentation URL would be used if the user triggers a presentation through the UA rather than through the page.

Is it intended that the default presentation URL is *also* used to filter devices for the purpose or the availabilitychange event ?

mfoltzgoogle:
> Is it intended that the default presentation URL is also used to filter devices for the purpose or the availabilitychange event ?

Yes, @avayvod and I believe it makes sense to do so.  The page has told us what content it wishes to present, so we should event the page only when there are screens available that can show that content.

However, this does raise some privacy concerns so it will need further discussion.

See https://github.com/w3c/presentation-api/pull/87

mfoltzgoogle:
PR #146 and the follow-on commit 586844704b025c416215dccb340f0d7229e17a4e describe the browser's behavior to setting `navigator.presentation.defaultRequest` in http://w3c.github.io/presentation-api/#setting-the-default-presentation-request.

There is an open issue relating to starting a session based on the default request. If the user initiates presentation and has pre-selected a screen, the UA doesn't need to ask the user again in http://w3c.github.io/presentation-api/#dfn-start-a-presentation-session. 

I'm leaving this open until the spec has been adjusted to address that issue.

mfoltzgoogle:
PR #274 adds some text for the case discussed in the previous note.

After some thinking I didn't think it was worth patching the start algorithm to explicitly bypass the "ask permission" step.  As written, I think there is leeway for the user agent to consider the pre-selection of the display (e.g., via tap-to-present) as granting permission, and make its own security/convenience tradeoff.  

I'll update the PR to make this more explicit.

mfoltzgoogle:
PR #274 merged.

PR #146, opened 2015-07-20
Introduce PresentationRequest and defaultRequest.

### LINKED ISSUE #138: Remove getAvailability() url argument unless implemented
This method was mentioned in https://lists.w3.org/Archives/Public/public-secondscreen/2015Jul/0002.html

> The group notes that a generic true/false is not enough in many 
> situations where the URL to present may require additional capabilities. 
> That is the reason why the "getAvailability" function takes the URL to 
> present as parameter, to allow the user agent to filter screens.

As actually implemented in Blink, the method takes an argument but it's never used. Unless there is an implementation that intends to do something with the argument, it would be better to not have it. Otherwise there will be no way to feature detect when it's *actually* supported, as `Presentation.prototype.getAvailability.length` will already be 1.

Slightly off-topic, it's not clear to me what could actually be done with just a URL. Presumably one would have to actually try to load the URL, make a guess about what features it needs to render, and then match that against the capabilities of the platform?

mounirlamouri:
Closes #138, #86 and #26.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,7 +9,7 @@
     "remove">
     </script>
     <script class="remove">
-      var respecConfig = {
+  var respecConfig = {
         specStatus: 'ED',
         edDraftURI: 'http://w3c.github.io/presentation-api/',
         shortName:  'presentation-api',
@@ -80,7 +80,7 @@
       };
     </script>
     <style>
-    /* Note formatting taken from HTML5 spec */
+/* Note formatting taken from HTML5 spec */
     .note { border-left-style: solid; border-left-width: 0.25em; background: none repeat scroll 0 0 #E9FBE9; border-color: #52E052; }
     .note em, .warning em, .note i, .warning i { font-style: normal; }
     p.note, div.note { padding: 0.5em 2em; }
@@ -494,10 +494,11 @@
           "controlling browsing context|controlling browsing contexts">controlling
           browsing context</dfn> (or <dfn>controller</dfn> for short) is a
           <a>browsing context</a> that has initiated or resumed a
-          <a>presentation session</a> by calling
-          <code><a for="PresentationRequest">start</a>()</code> or
-          <code><a for="PresentationRequest">join</a>()</code> or received a <a>presentation
-          session</a> via <code><a for="Presentation">defaultRequest</a></code>.
+          <a>presentation session</a> by calling <code><a for=
+          "PresentationRequest">start</a>()</code> or <code><a for=
+          "PresentationRequest">join</a>()</code> or received a <a>presentation
+          session</a> via <code><a for=
+          "Presentation">defaultRequest</a></code>.
         </p>
         <p>
           The <dfn>presenting browsing context</dfn> (or <dfn title=
@@ -792,8 +793,8 @@
             The user agent MUST keep track of the <dfn>set of availability
             objects</dfn> requested through the <code><a for=
             "PresentationRequest">getAvailability</a>()</code> method. The
-            <a>set of availability objects</a> is represented as a set of tuples
-            <em>(A, availabilityUrl, P)</em>, initially empty, where:
+            <a>set of availability objects</a> is represented as a set of
+            tuples <em>(A, availabilityUrl, P)</em>, initially empty, where:
           </p>
           <ol>
             <li>
@@ -924,10 +925,10 @@
               A</em>.
             </li>
             <li>If the <a>set of availability objects</a> is now empty and
-            there is no pending request to <a for="PresentationRequest" title=
-            "start">start a session</a>, cancel any pending task to <a>
-              monitor the list of available presentation displays</a> for power
-              saving purposes.
+            there is no pending request to <a for="PresentationRequest"
+              title="start">start a session</a>, cancel any pending task to
+              <a>monitor the list of available presentation displays</a> for
+              power saving purposes.
             </li>
           </ol>
           <p class="note">
@@ -950,20 +951,21 @@
 
             attribute EventHandler onconnect;
           };
-        </pre>
+
+</pre>
         <p>
           When a <a><code>PresentationRequest</code></a> is constructed, the
           given <code>url</code> MUST be used as the <dfn>presentation request
-          URL</dfn> which is the <a>presentation session URL</a> for the <a><code>PresentationRequest</code></a>
-          instance.
+          URL</dfn> which is the <a>presentation session URL</a> for the
+          <a><code>PresentationRequest</code></a> instance.
         </p>
         <section>
           <h4>
             Starting a presentation session
           </h4>
           <p>
-            When the <code><dfn for="PresentationRequest">start</dfn>()</code> method is
-            called, the user agent MUST run the following steps:
+            When the <code><dfn for="PresentationRequest">start</dfn>()</code>
+            method is called, the user agent MUST run the following steps:
           </p>
           <dl>
             <dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -366,12 +368,15 @@
...
   var presId = localStorage &amp;&amp; localStorage["presId"] || null;
-  // presId is mandatory for joinSession.
-  presId &amp;&amp; navigator.presentation.joinSession(presUrl, presId)
-    // The joined session will be passed to setSession on success
-    .then(setSession)
-    // no session found for presUrl and presId or an error is occurred
-    .catch(endSession);
+  // presId is mandatory when joining a session.
+  if (presId) {
+    var request = new PresentationRequest(presUrl);
```
avayvod: I think in the examples, we are reusing all the definitions from the previous ones, so there's no need to create a new PresentationRequest in each example. It might confuse developers, who will create a separate request availability, starting and joining presentation, etc.

index.html
```diff
@@ -382,12 +387,13 @@
         <pre class="example">
 &lt;!-- controller.html --&gt;
 &lt;head&gt;
-  &lt;!-- the link element with rel='default-presentation' allows the page to specify --&gt;
-  &lt;!-- the presentation URL and id for when the UA initiates a presentation session --&gt;
-  &lt;link href="http://example.com/presentation.html" rel="default-presentation" &gt;
+  &lt;!-- Setting presentation.defaultResquest allows the page to specify the
```
avayvod: Typo: s/Resquest/Request

index.html
```diff
@@ -382,12 +387,13 @@
...
+       default PresentationRequest to use when the UA initiates a presentation
```
avayvod: s/default//

index.html
```diff
@@ -1328,6 +1325,114 @@
...
+          <p>
+            An event named <code>connect</code> is fired on a
+            <a>PresentationRequest</a> when a session associated with the
+            object is initiated. It is fired at the <a>PresentationRequest</a>
+            instance, using the <a>PresentationSessionConnectEvent</a>
+            interface, with the <a for=
+            "PresentationSessionConnectEvent">session</a> attribute set to the
+            <a><code>PresentationSession</code></a> object that was initiated.
+            The event is fired regardless of how the session was initiated. For
+            example, it can come from a call to <code>start()</code>,
```
avayvod: Should we change/add the examples for that? E.g. not having start/join().then(setSession()) or provide an alternative version?

mwatson2: We should make it clear that the connect event is fired as soon as possible, i.e. as soon as the user selects a device, so that the page can update appropriately. For example, if the page is playing a video it would want to stop that immediately and indicate to the user that their request to fling the video to the device is in progress.

index.html

index.html

index.html

index.html
mounirlamouri: I think we should fire it as soon as there is a PresentationSession available, yes. Wrote a quick update.

mounirlamouri:
Please take another look.

index.html
```diff
@@ -332,8 +333,9 @@
     availability.onchange = function() { handleAvailabilityChange(this.value); };
   }.catch(function() {
     // Availability monitoring is not supported by the platform, discovery of presentation
-    // displays will happen only after startSession() call. Pretend the devices are
-    // available for simplicity (one could implement the third state for the button).
+    // displays will happen only after PresentationRequest.start() call. Pretend the
```
mfoltzgoogle: after request.start()

index.html
```diff
@@ -382,12 +384,12 @@
         <pre class="example">
 &lt;!-- controller.html --&gt;
 &lt;head&gt;
-  &lt;!-- the link element with rel='default-presentation' allows the page to specify --&gt;
-  &lt;!-- the presentation URL and id for when the UA initiates a presentation session --&gt;
-  &lt;link href="http://example.com/presentation.html" rel="default-presentation" &gt;
+  &lt;!-- Setting presentation.defaultRequest allows the page to specify the
+       PresentationRequest to use when the UA initiates a presentation session --&gt;
```
mfoltzgoogle: Add punctuation at the end?

index.html
```diff
@@ -382,12 +384,12 @@
...
 &lt;/head&gt;
 &lt;script&gt;
-  navigator.presentation.ondefaultsessionstart = function (evt) {
+  navigator.presentation.defaultRequest = new PresentationRequest(defaultUrl);
+  navigator.presentation.defaultRequest.onsessionconnect = function(evt) {
```
mfoltzgoogle: We use the `sessionavailable` event on NavigatorPresentation to indicate when a session has connected to the presented page [1].  It would seem more economical to use the same event to indicate that the controlling page has connected to the presented page and vice versa.  At first glance I prefer the name `sessionconnect` for both; I will file a followup issue.

[1] https://w3c.github.io/presentation-api/#monitoring-incoming-presentation-sessions-in-a-presenting-browsing-context

index.html
```diff
@@ -483,15 +485,16 @@
...
           <a>browsing context</a> that has initiated or resumed a
-          <a>presentation session</a> by calling
-          <code><a>startSession</a>()</code> or
-          <code><a>joinSession</a>()</code> or received a <a>presentation
-          session</a> via <code><a>ondefaultsessionstart</a></code> event.
+          <a>presentation session</a> by calling <code><a for=
+          "PresentationRequest">start</a>()</code> or <code><a for=
+          "PresentationRequest">join</a>()</code> or received a <a>presentation
+          session</a> via <code><a for=
+          "Presentation">defaultRequest</a></code>.
```
mfoltzgoogle: I would use two sentences here, one for the page-initiated case and the other for the UA-initiated case.

index.html
```diff
@@ -1328,6 +1321,118 @@
...
+
+            dictionary PresentationSessionConnectEventInit : EventInit {
+              PresentationSession session;
+            };
+
+</pre>
+          <p>
+            An event named <code>sessionconnect</code> is fired on a
+            <a>PresentationRequest</a> when a session associated with the
+            object is initiated. It is fired at the <a>PresentationRequest</a>
```
mfoltzgoogle: s/initiated/connected/ here and below

index.html
```diff
@@ -1328,6 +1321,118 @@
...
+            "PresentationSessionConnectEvent">session</a> attribute set to the
+            <a><code>PresentationSession</code></a> object that was initiated.
+            The event is fired regardless of how the session was initiated. For
+            example, it can come from a call to <code>start()</code>,
+            <code>join()</code> or even from a UA initiated session on <a for=
+            "Presentation"><code>defaultRequest</code></a>.
+          </p>
+          <p>
+            The UA MUST fire the event as soon as it can create the
+            <a><code>PresantiotSession</code></a> associated to the event.
```
mfoltzgoogle: Typo in PresentationSession

index.html
mfoltzgoogle: The start and join algorithms need to both be updated to fire this new event.  I believe that should handle the use case for @mwatson2, since we don't block resolution of the Promise on connection of the session.

The example code should also be updated to reflect that the session may initially be disconnected.

index.html
```diff
@@ -1328,6 +1321,118 @@
...
+          otherwise.
+        </p>
+        <section>
+          <h4>
+            Default presentation request
+          </h4>
+          <p>
+            If set, the <a for="Presentation">defaultRequest</a> SHOULD be used
+            by the UA as the <dfn>default presentation request</dfn>. When the
+            UA wants to start or join a session from its UI, it MUST use the
```
mfoltzgoogle: I would prefer this as SHOULD for two reasons:
  - There are use cases for the user or UA overriding this behavior.
  - On some UAs with limitations on the user interface, it may not be possible to initiate a presentation from the UA at all.

mfoltzgoogle:
Thanks for the contribution @mounirlamouri and the comments @avayvod and @mwatson2. I made some inline comments which I can implement in a followup PR.

PR #147, opened 2015-07-21
Fetch respec over HTTPS

foolip:
This ought to fix mixed content problems on https://w3c.github.io/presentation-api/

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,7 @@
     <title>
       Presentation API
     </title>
-    <script src="http://www.w3.org/Tools/respec/respec-w3c-common" async class=
+    <script src="https://www.w3.org/Tools/respec/respec-w3c-common" async class=
     "remove">
     </script>
     <script class="remove">


mounirlamouri:
I can't merge but LGTM.

tidoust:
@foolip I'm afraid I broke this while switching to ReSpec. Sorry about that!

Could we rather use a "network-path reference", meaning dropping the `https:` prefix to end up with `//www.w3.org/Tools/respec/respec-w3c-common` so that it works with both schemes?

foolip:
WFM, but I was assuming that people want to edit the file locally, in which case the scheme-relative URL wouldn't work.

mounirlamouri:
Given that it's fine to include https from http, I think we can go with
https only.

foolip:
Thanks, https://w3c.github.io/presentation-api/ now renders as it should!

PR #156, opened 2015-08-16
Use「Living Document」convention for auto-publishing

sideshowbarker:
Use **Living Document** subtitle to indicate in the TR/WD version that it’s up-to-date with the latest repo sources (auto-published).

I think @marcoscaceres introduced this convention with the https://github.com/w3c/manifest spec, but regardless it seems like a useful convention to adopt for specs that are using the auto-publishing system.

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,6 +13,7 @@
         specStatus: 'ED',
         edDraftURI: 'http://w3c.github.io/presentation-api/',
         shortName:  'presentation-api',
+        subtitle:  'Living Document',
         editors: [
           {
             w3cid: 68454,


anssiko:
Thanks for the PR. The group agreed to use a semi-automatic approach with Echidna (over auto-publishing), where the TR/WD is updated at editor's discretion and not on every commit automatically. This is documented in the [work mode wiki][1]. So it would be probably a bit misleading to use the awesome Living Document subtitle for this spec currently. Should the group opt-in for auto-publishing in the future, we'd definitely adopt the subtitle.

@sideshowbarker If you agree, we'd close this PR.

[1]: https://www.w3.org/wiki/Second_Screen/Work_Mode#Publication_as_Working_Draft

sideshowbarker:
> The group agreed to use a semi-automatic approach with Echidna (over auto-publishing), where the TR/WD is updated at editor's discretion and not on every commit automatically. 
> @sideshowbarker If you agree, we'd close this PR.

Yep, agreed, so I’m going ahead and closing it now

PR #157, opened 2015-08-17
Use syntax highlighting in the examples

anssiko:
Make use of the [syntax highlighter][1] of ReSpec.

[1]: https://www.w3.org/respec/ref.html#highlight


mfoltzgoogle:
LGTM.  Thanks @anssiko

Issue #35, opened 2014-12-18
Refine how to do session teardown/disconnect/closing


drott:
There are rough edges when it comes to closing the session, resuming it. We need to gain experience from implementations what happens, when the session is disconnected, user-closed, remote-end closed etc.

This section is one aspect of this, but probably the example section needs to shed more light on how expected and unexpected disconnects are handled.

drott:
https://lists.w3.org/Archives/Public/public-secondscreen/2015Jan/0012.html (End of the mail "semantics of close"

mfoltzgoogle:
Copied from #33:

Presumably the user agent will provide the ability for the browser with the initiating page to close the presented document (local or remote). However we may want the user agent to close it on the user's behalf when:

1. There are no controlling sessions connected for some period of time (e.g., idle timeout)
2. The presentation is being rendered locally (1-UA) and the media stream has has been terminated permanently.
3. The presentation is being rendered locally and the user wishes to exit the user-agent.
4. The presentation wishes to terminate itself.

Some of these scenarios can be implemented by the app developer; if so the behavior of e.g. window.close() should be defined in the presentation context.

schien:
I'd like to put my two cents before the F2F meeting.

Currently `PresentationSessionState` only support two states: `connected` and `disconnected`, which is not enough for both requesting page and presenting page to distinguish between terminating the presentation or presentation display being disconnected accidentally.

This can be solved by introducing additional state `terminated`. While `session.close()` (or rename to `session.terminate()`) being invoked, `session.state` will enter `terminated` state. Application can reclaim the resources and terminate itself, e.g. closing the window.

mfoltzgoogle:
Thanks, that seems reasonable.  Other options that I would put forth for consideration:

1. Adding a `disconnectReason` property to the `statechange` event that lets the page know why the `PresentationSession` was disconnected.
2. Adding an error event (and handler) to `PresentationSession.`  If the session was terminated accidentally, this would fire.  This is similar to `RTCDataChannel` [1], although the semantics of `RTCDataChannel.onerror` are not clear from that spec.

[1] http://www.w3.org/TR/webrtc/#rtcdatachannel

matt-hammond-bbc:
Some underlying protocols may not support terminating/closing the 2nd screen content. In DIAL, the ability to "stop" an application is optional [1]. Similarly HbbTV (that builds ontop of DIAL) does not mandate this functionality.

[...]

Issue #80, opened 2015-04-24
Define security requirements for messaging channel between secure origins


mfoltzgoogle:
[public-secondscreen][1] thread:

While discussing the [Hbb 2.0 TV standard](http://www.digitaltveurope.net/320392/hbbtv-2-0-specification-announced/), the issue of how WebSockets would be used to communicate between the controlling Web application and the presenting TV Web application.

If these Web applications are provided on secure (`https://`) origins, some guarantees of message confidentiality and authenticity of either party should be made to meet the standards set out by the [Mixed Content proposal](https://w3c.github.io/webappsec/specs/mixedcontent/).

This issue will be addressed by a spec change to spell out the security requirements for the messaging channel as part of the Security and Privacy sections of the spec.

[1]: https://lists.w3.org/Archives/Public/public-secondscreen/2015Apr/0055.html

anssiko:
Tagged this for F2F discussion. @mfoltzgoogle mentioned having had "preliminary discussions in Chrome about this, but nothing firm enough to propose, so I'm glad to have an opportunity to discuss this in the group."

So let's discuss at F2F latest :-)

tidoust:
PROPOSED RESOLUTION: keep issue 80 open while we gather more implementation experience. Highlight issue asking for feedback when getting in touch with PING / Web security IG

See relevant discussion during first day of Berlin F2F:
http://www.w3.org/2015/05/19-webscreens-minutes.html#item09

mfoltzgoogle:
Re-adding F2F tag so that we can review feedback from PING/TAG and reports from implementations in October.

mwatson2:
We would like to propose some ideas for discussion on this topic at the F2F (actually, more leaning towards possible solutions than just requirements). Might also be worth a session in the Unconference part of TPAC to get wider visibility of the problem. So, I support Mark's addition of the F2F tag.

Btw, has there been any TAG discussion of this ?

tidoust:
The TAG prepared a draft review of the Presentation API. The relevant part appears more descriptive than prescriptive:

[[
[...]

Issue #14, opened 2014-09-03
Define user agent context for rendering the presentation


mfoltzgoogle:
The spec should specify any restrictions on the presenting browsing context when the opening browsing context is in "incognito" mode.

From [public-webscreens][1] and related threads:

The specification should define the behavior of the following Web features for the presenting page across presentation sessions. The goals are the following:

1. Ensure that the same presentation will behave correctly in the 1-UA and 2-UA cases
1. Information does not leak from presentation session to session in cases where the controlling pages may represent different authorization domains (e.g., logged-in users).

The following Web features have been discussed (and there may be others to consider):
* Cookies and local storage
* Sensor APIs (geo, orientation, etc.)
* Browser extensions
* HTTP Cache
* Shared Workers

[1]: https://lists.w3.org/Archives/Public/public-webscreens/2014Aug/0012.html

tidoust:
See related discussion and action at Berlin F2F:
http://www.w3.org/2015/05/19-webscreens-minutes.html#item13

anssiko:
ACTION: @mfoltzgoogle to define browsing context in terms of the upcoming spec for private browsing, perhaps using the Mozilla link as an interim reference. [recorded in http://www.w3.org/2015/05/19-webscreens-minutes.html#action14]

anssiko:
Should also decide whether the presenting page is able to navigate to another URL.

Issue #10, opened 2014-09-03
Is user permission required to prompt for screen availability information?


mfoltzgoogle:
We do not want to require user permission before disclosing the presence of a presentation display, as it is counter to the initial purpose of improving the user experience.

jcdufourd:
If the page displaying this info is the result of a direct user interaction, then I would think not.
If the page is started by a script and not rendered, then yes.
So I guess my answer is, it depends.

mounirlamouri:
Closing in favour of working on PING feedback.

PR #158, opened 2015-08-17
Pull in issue tracker data using GitHub API

anssiko:
ReSpec has a feature that allows to pull in data using the GitHub API. This PR makes use of that feature and pulls open issue data directly from GH issues into the spec. This change should make it easier for us to keep the spec and GH issues in sync, as the latter will be the canonical source for all issue data. Any changes done to the GH issues (specifically, its topic or the first comment) will be automatically pulled into the latest spec.

For example, the following markup in the ReSpec source:

    <div class="issue" data-number="35"></div>

Is turned into the following markup automatically:

    <div class="issue" id="issue-35">
      <div class="issue-title" aria-level="5" role="heading" id="h-issue1">
        <a href="https://www.github.com/w3c/presentation-api/issues/35">
        <span>ISSUE 35</span></a>:    
        Refine how to do session teardown/disconnect/closing
      </div>
        <p>
          There are rough edges [... first comment in the respective GH issue ...].
        </p>
    </div>

@mfoltzgoogle Please review and merge if you're happy with the change.

### DIFF
--- a/index.html
+++ b/index.html
@@ -76,7 +76,9 @@
             status: 'finding',
             publisher: 'W3C'
           }
-        }
+        },
+        issueBase: "https://www.github.com/w3c/presentation-api/issues/",
+        githubAPI: "https://api.github.com/repos/w3c/presentation-api"
       };
     </script>
     <style>
@@ -88,14 +90,6 @@
     .note p:first-child { margin-top: 0; }
     .note p:last-child { margin-bottom: 0; }
     p.note:before { content: 'NOTE: '; }
-
-    .open-issue { border-left-style: solid; border-left-width: 0.25em; background: none repeat scroll 0 0 #fbfbe9; border-color: #faf9a5; }
-    .open-issue em, .warning em, .open-issue i, .warning i { font-style: normal; }
-    p.open-issue, div.open-issue { padding: 0.5em 2em; }
-    span.open-issue { padding: 0 2em; }
-    .open-issue p:first-child { margin-top: 0; }
-    .open-issue p:last-child { margin-bottom: 0; }
-
     .non-normative { border-left-style: solid; border-left-width: 0.25em; background: none repeat scroll 0 0 #E9FBE9; border-color: #52E052; }
     p.non-normative:before { content: 'Non-normative: '; font-weight: bolder;}
     p.non-normative, div.non-normative { padding: 0.5em 2em; }
@@ -571,7 +565,7 @@
           <h4>
             Sending a message through <code>PresentationSession</code>
           </h4>
-          <p class="note">
+          <div class="note">
             No specific transport for the connection between the <a>controlling
             browsing context</a> and the <a>presenting browsing context</a> is
             mandated, except that for multiple calls to <code><a for=
@@ -579,7 +573,7 @@
             messages are delivered to the other end reliably and in sequence.
             The transport should function equivalently to an
             <a><code>RTCDataChannel</code></a> in reliable mode.
-          </p>
+          </div>
           <p>
             Let <dfn>presentation message data</dfn> be the payload data to be
             transmitted between two browsing contexts. Let <dfn>presentation
@@ -707,10 +701,7 @@
               </ol>
             </li>
           </ol>
-          <p class="open-issue">
-            <a href="https://github.com/w3c/presentation-api/issues/35">ISSUE
-            35: Refine how to do session teardown/disconnect/closing</a>
-          </p>
+          <div class="issue" data-number="35"></div>
         </section>
         <section>
           <h4>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1651,21 +1634,8 @@
...
-        presented?</a>
-      </p>
-      <p class="open-issue">
-        We do not want to require user permission before disclosing the
-        presence of a presentation display, as it is counter to the initial
-        purpose of improving the user experience. See <a href=
-        "https://github.com/w3c/presentation-api/issues/10">ISSUE 10: Is user
-        permission required to prompt for screen availability information?</a>
-      </p>
+      <div class="issue" data-number="9"></div>
```
mounirlamouri: Did you remove the text on purpose?

index.html
```diff
@@ -1651,21 +1634,8 @@
...
+      <div class="issue" data-number="10"></div>
```
mounirlamouri: ditto

index.html
```diff
@@ -1731,16 +1701,11 @@
...
         the user requests to "clear browsing data."
       </p>
-      <p class="open-issue">
-        The spec should specify any restrictions on the presenting browsing
-        context when the opening browsing context is in "incognito" mode. See
-        <a href="https://github.com/w3c/presentation-api/issues/14">ISSUE 14:
-        Define user agent context for rendering the presentation</a>
-      </p>
-      <p class="open-issue">
+      <div class="issue" data-number="14"></div>
```
mounirlamouri: ditto

index.html
```diff
@@ -1751,11 +1716,7 @@
         confidentiality and authenticity between corresponding <a>presentation
         sessions</a>.
       </p>
-      <p class="open-issue">
-        <a href="https://github.com/w3c/presentation-api/issues/80">ISSUE 80:
-        Define security requirements for messaging channel between secure
-        origins</a>
-      </p>
+      <div class="issue" data-number="80"></div>
```
mounirlamouri: ditto

index.html
anssiko: https://github.com/w3c/presentation-api/issues/9 should have the same text.

index.html
anssiko: https://github.com/w3c/presentation-api/issues/10

index.html
anssiko: Updated https://github.com/w3c/presentation-api/issues/14 with the removed text.

index.html
anssiko: https://github.com/w3c/presentation-api/issues/80

Please note the spec is rendered as follows https://rawgit.com/anssiko/presentation-api/issue-tracker-integration/index.html with data pulled from GH issues using GitHub API.

mounirlamouri:
Thanks for the generated HTML. The issue summary looks a bit too verbose but I guess it will be fixed if we actually fix the issues ;)

LGTM

anssiko:
That was the idea -- by making the issues more visible in the spec we put a bit of pressure on the group to fix them :-)

We can amend and embellish the issue summaries as we roll on, make them prettier.

mfoltzgoogle:
I am not sure this is a positive change.  It appears that the effective change is to pull the first comment from the GH issue into the document, which could be quite long and out of context.  As is, the change hurts readability.

I would be okay with this approach with two modifications:
- Change the integration to omit the summaries and only include the issue title.
- Have ReSpec omit the issue when the issue is closed on GH.

If these aren't possible, I would like to revert this change to the prior approach, where the issues are just links to GH and the reader can see the entire issue thread in context.

anssiko:
@mfoltzgoogle Would your first concern be addressed if the first comment of each issue that is pulled in the spec would be edited in GH issues to be worded as such that it is fine for inclusion into the spec while further elaboration would be placed in the follow up comments? We've used this integration in some other specs and I personally felt keeping GH issues in sync with the spec makes the editor's life easier since it minimizes information duplication. The only tradeoff I noticed is that since this information is pulled from GH into the spec dynamically the source itself does not include the issue text, source diff for issue text changes is not possible unless we generate an HTML from the source on each commit (we do that with ReSpec for some specs).

I think the second concern is already addressed i.e. closed issues are not rendered, but it would probably still make sense to drop the issue placeholder `<div>`s from the specification when the corresponding issue is closed similarly to as we'd do even without this integration.

If the above does not address the concerns and  you feel this integration is not actually helping you as an editor feel free to revert. It is always the editor who has the final say on the toolchain issues :-)

mfoltzgoogle:
I'm not sure, if we edited every referenced issue as described wouldn't we lose some information (or lose the correct chronological sequence of comments)?

Auto-removal of closed issues is nice, but since the initial comment in a GH issue rarely changes, I'm not sure there's a lot of synchronization required.

anssiko:
@mfoltzgoogle I submitted https://github.com/w3c/presentation-api/pull/174 to address your first concern.

PR #159, opened 2015-08-17
PresentationRequest.start() requires a user gesture.

mounirlamouri:
Like most features showing a prompt, it should be bound to a user gesture, otherwise this is going to lead to very confusing UX.

@avayvod @anssiko @mfoltzgoogle PTAL

### DIFF
--- a/index.html
+++ b/index.html
@@ -257,7 +257,9 @@
         <dfn><a href="http://www.w3.org/TR/html5/webappapis.html">queue a
         task</a></dfn>, <dfn><a href=
         "http://www.w3.org/TR/html5/infrastructure.html#concept-events-trusted">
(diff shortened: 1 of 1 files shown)


avayvod:
lgtm

do you have a demo link?

anssiko:
Thanks. This change is consistent with the behaviour of other features that prompt the user. For example, `<input type=file>` and `<a download>`.

With this change, the `start()` method provides more fine-grained information on the reason of rejection. This should allow the web developer to come up with a fallback that is more specific and/or inform the user of the exact reason of the failure.

LGTM.

mounirlamouri:
Anssi, any chance you can merge that? Mark is on holidays this week and neither Anton or I have edit access.

Issue #164, opened 2015-08-21
1-UA and 2-UA details


hsablonniere:
Hi everyone ;-)

In many places, we can find references to 1-UA and 2-UA ([Chromium Intent to Ship](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/YKVnS429hns), Firefox bugzilla [#1184036](https://bugzilla.mozilla.org/show_bug.cgi?id=1184036) & [#1184073](https://bugzilla.mozilla.org/show_bug.cgi?id=1184073), spec Github issues...) but I cannot find a formal definition in the spec.

Here's what I understood :
* 1-UA is for non Web capable presentation displays, the UA initiating the request is rendering the page himself and streams it to another presentation display.
* 2-UA is for Web browser capable presentation displays such as the Chromecast, Matchstick...

The description in Bugzilla [#1184036](https://bugzilla.mozilla.org/show_bug.cgi?id=1184036) regarding 1-UA mentions the Chromecast. It confuses me. I thought it was a 2-UA presentation display.

In the Chromium Intent to Ship post, it says that 1-UA is not implemented yet but it will be before the end of the year. I'm confused when they say they will implement 1-UA even if browser vendors are free to do so. I can't find where the spec mentions that.

Thanks.

anssiko:
Thanks for the feedback. Perhaps Chromium participants in this group have some language in their design doc that could be reused to clarify this in the introduction section. I can confirm that this terminology has caused some confusion around people not following the work closely and we should clarify this. Not a blocker, so assigning P3.

hsablonniere:
Thx

hsablonniere:
Thanks @anssiko.

If found this in the charter :
> Content mirroring, whereby a web application requests display of the content shown in its own browsing context (i.e., page) on a secondary display, is out of scope.

But it's about mirroring which seems to be different from 1-UA.

tidoust:
The 1-UA and 2-UA possibilities are described in the spec introduction, except they are not given specific names there. Given that we keep referring to them as 1-UA and 2-UA in the group and elsewhere, it makes sense to mention these terms in the spec. One quick and easy fix would be to add something along the lines of "(commonly referred to as the 1-UA case)" next to the relevant bits.

For reference, the relevant bits are:

[...]

PR #166, opened 2015-08-21
Mention 1-UA and 2-UA cases explicitly in the introduction

hsablonniere:
See #164 for details.


anssiko:
@hsablonniere thanks for the clarification of the 1-UA and 2-UA terminology in the introduction.

mfoltzgoogle:
LGTM Thanks again @hsablonniere.

PR #169, opened 2015-08-22
Use TIDY instead of TIDY-HTML5 to make it settable in from ENV.

mounirlamouri:
This said, I'm not entirely sure why we use ```tidy-html5``` as the name of the binary while ```tidy``` is the output of building the tool from the repository. Am I doing something wrong?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-TIDY-HTML5 ?= tidy-html5
+TIDY ?= tidy-html5
 
 tidy: index.html build/tidyconfig.txt
-	$(TIDY-HTML5) -config build/tidyconfig.txt -o $< $<
+	$(TIDY) -config build/tidyconfig.txt -o $< $<


tidoust:
I believe that the tidy repository used to generate a `tidy-html5` binary (perhaps as a way to disambiguate with the good ol' tidy for HTML4 tool). That is no longer the case so it makes sense to me to switch both the ENV variable and the default binary name to just `tidy`.

anssiko:
This is a historical artefact. I suggest we modernize the `Makefile` and update https://www.w3.org/wiki/Second_Screen/Work_Mode accordingly. Feel free to work on this.

Thanks, I'll merge this PR.

mounirlamouri:
I've updated the doc to mention ```TIDY=tidy make```.

mfoltzgoogle:
LGTM.  Thanks @mounirlamouri

PR #172, opened 2015-08-26
getAvailability() should return true if the list of displays isn't empty.

### LINKED ISSUE #161: Let getAvailability() resolve its promise as soon as availability has been tested
http://w3c.github.io/presentation-api/#dom-presentationrequest-getavailability

Reliable sources (@avayvod) tell me that the implementation in Chromium "will resolve the promise with availability false and then update the value with true as soon as the first device is found" which isn't what the spec says.

The exact details are editorial, but I would propose to resolve the promise within the definition of getAvailability() and to let the monitoring steps only fire events in case of changes.

mounirlamouri:
In other words, if there is a discovery happening which already found a device, getAvailability() should be resolved with true.

Fixes #161

### DIFF
--- a/index.html
+++ b/index.html
@@ -822,7 +822,7 @@
             when there are available displays. However, the user agent may not
             support continuous availability monitoring; for example, because of
             platform or power consumption restrictions. In this case the
-            <a>Promise</a> returned by <code>getAvailability()</code> is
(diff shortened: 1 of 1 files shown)


avayvod:
lgtm

mfoltzgoogle:
LGTM. Thanks @mounirlamouri

PR #173, opened 2015-08-26
getAvailability: resolve with false if UA can't even start a session.

mounirlamouri:


index.html
```diff
@@ -1291,8 +1291,22 @@
...
+                  object with its <code>value</code> property set to <code>false</code>.
+                </li>
+                <li>
+                  Abort all the remaining steps.
+                </li>
+              </ol>
+            </li>
             <li>If the user agent is unable to continuously <a>monitor the list
-            of available presentation displays</a>, then:
+            of available presentation displays</a> but there might be available
```
mfoltzgoogle: ... but may be able to resume monitoring at a later time, then:

index.html
```diff
@@ -1291,8 +1291,22 @@
             </li>
             <li>Return <em>P</em>.
             </li>
+            <li>If the user agent knows there will be no available presentation
```
mfoltzgoogle: If the user agent is unable to monitor presentation displays for the entire duration of the controlling browsing context (e.g., because the user has disabled this feature), then:

mfoltzgoogle:
Looks okay, I proposed some alternative wording.

index.html
mounirlamouri: I don't think being able to monitor later should be added. It might be very possible to not be able to monitor at all but still be able to connect to a device, right?

index.html

mounirlamouri:
Comments applied. PTAL.

index.html
```diff
@@ -1319,7 +1333,8 @@
             <li>Let <em>A</em> be a new <code>PresentationAvailability</code>
             object with its <code>value</code> property set to
             <code>false</code> if the <a>list of available presentation
-            displays</a> is empty and <code>true</code> otherwise.
+            displays</a> is empty or non of them is a <a>compatible
+            presentation display</a>, <code>true</code> otherwise.
             </li>
             <li>Create a tuple <em>(A, presentationUrl, P)</em> and add it to
```
mfoltzgoogle: Since we are always resolving P we may not need to keep it around to resolve it later; perhaps P can be dropped from the set of availability objects?  This cleanup can be made later.

mfoltzgoogle:
LGTM with one comment.

PR #174, opened 2015-08-27
Hide issue summaries pulled from GH

anssiko:
@mfoltzgoogle This is a workaround to hide the (currently too verbose) issue summaries pulled from GH. If this feature is liked by others we should bake this into ReSpec as a configuration option. In the interim, this CSS should do.


mfoltzgoogle:
Some fancy CSS to makes the editor happy :)
Thanks @anssiko

PR #175, opened 2015-08-27
Initial version of the interoperability document

louaybassbouss:
Initial version of the interoperability document according to the discussion in issue #153.

PR #176, opened 2015-08-27
Editorial

anssiko:
Add anchors for 1-UA and 2-UA terms to allow cross-referencing from interoperability.md that is being baked in https://github.com/w3c/presentation-api/pull/175.


mounirlamouri:
lgtm, thanks!

PR #179, opened 2015-08-31
Resolve getAvailability with current status.

mounirlamouri:
Resolving the getAvailability() call with the currently known status. Thanks to @foolip for catching the mistake.


foolip:
Looks good now, thanks!

PR #180, opened 2015-08-31
(cleanup) set of availability objects: remove P from the tuple.

mounirlamouri:


mfoltzgoogle:
Thanks for the fast turnaround @mounirlamouri!

PR #181, opened 2015-08-31
Rename PresentationSession.join() to reconnect().

mounirlamouri:
Fixes #170.

PR #182, opened 2015-08-31
Editorial: fix the the spec warnings.

mounirlamouri:
This is basically changing "title=" to "data-lt=".


mfoltzgoogle:
LG with comments applied from PR #181.

PR #185, opened 2015-09-07
Make session a required argument when constructing a PresentationSessionConnect.

### LINKED ISSUE #184: PresentationSessionConnectEventInit.session should be either [required] or nullable (and default to null)
...otherwise it isn't clear what
(new PresentationSessionConnectEvent("")).session does.

If we go with nullable session in the dictionary, also the session property on the event should be nullable.

mounirlamouri:
Fixes #184.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1396,13 +1396,13 @@
             Interface <code>PresentationSessionConnectEvent</code>
           </h3>
           <pre class="idl">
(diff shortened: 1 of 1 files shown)


mounirlamouri:
@smaug---- WDYT?

mfoltzgoogle:
LGTM

Issue #93, opened 2015-05-26
Conformance classes definition


tidoust:
The Presentation API distinguishes between the two sides of the equation: the *opening browsing context* and the *presenting browsing context*. The interfaces that user agents need to implement are different depending on which side you consider. Also, I think we envision user agents that can run an opening browsing context but cannot run a presenting browsing context, and conversely user agents that can run a presenting browsing context but cannot run an opening browsing context.

In spec parlance, this means that there exists (at least) two different classes of products that can claim conformance against the Presentation API specification. The specification should be explicit about this in the Conformance section and clearly identify the interfaces, algorithms and normative statements that apply to a given class of products.

This is somewhat different from most other JavaScript APIs that target a single user agent and thus can remain pretty laconic on the definition of classes of products. As a counter example, the HTML5 specification defines different classes of products, including *Conforming documents*, *Web browsers and other interactive user agents* and *Conformance checkers* [html5].

The definition of classes of products can remain pretty simple. As a starting point, I would suggest to reference the definitions of *opening browsing context* and *presenting browsing context* with wording in the Conformance section such as:

[[
This specification describes conformance criteria for two classes of products:

 * An **opening user agent** is a user agent that can initiate or resume a presentation session, thus creating an *opening browsing context*. An opening user agent must implement the interfaces defined in sections *(to be completed)*.
 * A **presenting user agent** is a user agent that can run a *presenting browsing context*. A presenting user agent must implement the interfaces defined in sections *(to be completed)*.

A user agent may act both as an *opening user agent* and as a *presenting user agent*, provided it implements all the required interfaces. This happens, e.g. when the opening user agent runs the *presenting browsing context* locally, or when the presenting user agent wants to allow a presenting browsing context to initiate or resume a presentation session on its own.
]]

We could also treat the 1UA case as a separate class of products but I am not yet clear whether we will need to define normative statements that will only apply to the presenting browsing context when it runs in the opening user agent.

Having clear classes of products is good practice (see relevant guideline in "QA Framework: Specification Guidelines" [qa]). It is useful to make conformance claims but note it also helps with testing (see "A Method for Writing Testable Conformance Requirements" [test]) and specification authoring (once you have clear classes of conformance, you will probably want to separate interfaces as suggested in #91 to make it easier for readers/developers).

Feel free to suggest better definitions. The main point in this issue is not the text in itself but the fact that the spec needs to define the conformance classes.

[...]

PR #186, opened 2015-09-08
Add PresentationReceiver and navigator.presentation.receiver to access it.

### LINKED ISSUE #91: Separate interface for controlling page and presenting page

mounirlamouri:
This is moving around some spec definitions to apply on
PresentationReceiver instead of Presentation, the examples have been
changed and navigator.presentation.receiver has been defined.

Fixes #91.

### DIFF
--- a/index.html
+++ b/index.html
@@ -447,6 +447,11 @@
 &lt;/script&gt;
 
 </pre>
+      </section>
+      <section>
+        <h3>
+          Presenting context: monitor available session(s) and say hello.
+        </h3>
         <pre class="example highlight">
 &lt;!-- presentation.html --&gt;
 &lt;script&gt;
@@ -460,9 +465,9 @@
         session.send("hello");
     }
   });
-  navigator.presentation.getSession().then(addSession);
-  navigator.presentation.onsessionavailable = function(evt) {
-    navigator.presentation.getSessions().then(function(sessions) {
+  navigator.presentation.receiver.getSession().then(addSession);
+  navigator.presentation.receiver.onsessionavailable = function(evt) {
+    navigator.presentation.receiver.getSessions().then(function(sessions) {
       addSession(sessions[sessions.length-1]);
     });
   };
@@ -1428,24 +1433,10 @@
       </section>
       <section>
         <h3>
-          Interface <a><code>Presentation</code></a>
+          Interface <a><code>PresentationReceiver</code></a>
         </h3>
         <pre class="idl">
-          partial interface Navigator {
-            [SameObject] readonly attribute Presentation presentation;
-          };
-
-</pre>
-        <p>
-          The <dfn for="Navigator"><code>presentation</code></dfn> attribute is
-          used to retrieve an instance of the <a>Presentation</a> interface.
-        </p>
-        <pre class="idl">
-          interface Presentation : EventTarget {
-            // This API used by controlling browsing context.
-            attribute PresentationRequest? defaultRequest;
-
-            // This API used by presenting browsing context.
+          interface PresentationReceiver : EventTarget {
             Promise&lt;PresentationSession&gt; getSession();
             Promise&lt;sequence&lt;PresentationSession&gt;&gt; getSessions();
             attribute EventHandler onsessionavailable;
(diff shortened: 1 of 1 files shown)


anssiko:
See https://github.com/w3c/presentation-api/issues/91#issuecomment-138535506 re the conformance section.

mounirlamouri:
I think my wording is fine. If a browsing context isn't a presenting browsing context, it MUST return null, and return an instance of PresentationReceiver otherwise. Clearly, that means that a controlling UA MUST implement PresentationReceiver.

What do you think should be added?

anssiko:
@tidoust Please review and see if this matches your expectations re conformance requirements.

tidoust:
@mounirlamouri Could you clarify why you want a controlling UA to implement PresentationReceiver? Or is it a typo in your comment?

I do not know in which order #91 and #93 should be done. I think that the pull request is fine provided that we keep #93 open. Once "presenting UA" and "controlling UA" are defined as conformance classes, sentences such as: "The receiver MUST return null if the current browsing context is not a presenting browsing context and return a PresentationReceiver instance otherwise" may need to be revisited to clarify what each type of UA needs to implement precisely.

For instance, here, I am not clear what "typeof PresentationRequest" will return in a presenting-only UA. If defined, that would mean that the UA needs to implement the Constructor too and thus some "fake" logic. How would a Web app know that it's useless to offer the user with a "present" button in that case?

mounirlamouri:
Indeed, that was a typo, I meant that a presenting UA MUST implement PresentationReceiver.

A controlling MAY implement PresentationReceiver. I guess the only question is whether a controlling UA should have the interface in its global scope or not. I don't think that's a major issue and it could probably be defined in #93.

index.html
```diff
@@ -447,6 +447,11 @@
 &lt;/script&gt;
 
 </pre>
+      </section>
+      <section>
+        <h3>
+          Receiver: monitor available session and say hello
```
mfoltzgoogle: Is "receiver" defined yet in the document?  Would restate as:

Presenting context: monitor available session(s) and say hello.

index.html
```diff
@@ -1453,45 +1444,13 @@
 
 </pre>
         <p>
-          The <dfn for="Presentation"><code>defaultRequest</code></dfn> MUST
-          return the <a>default presentation request</a> if any, null
-          otherwise.
+          The <a>PresentationReceiver</a> object is available to a <a>presenting
```
mfoltzgoogle: The spec should convert to use the terminology "receiving browsing context" to align API naming and API concepts.

index.html
```diff
@@ -1648,6 +1607,76 @@
...
+        <p>
+          The <dfn for="Navigator"><code>presentation</code></dfn> attribute is
+          used to retrieve an instance of the <a>Presentation</a> interface.
+        </p>
+        <pre class="idl">
+          interface Presentation {
+            // This API used by controlling browsing context.
+            attribute PresentationRequest? defaultRequest;
+
+            // This API is available on presenting browsing context.
```
mfoltzgoogle: ...available in the receiving browsing context.

index.html
```diff
@@ -1648,6 +1607,76 @@
...
+          };
+
+</pre>
+        <p>
+          The <dfn for="Navigator"><code>presentation</code></dfn> attribute is
+          used to retrieve an instance of the <a>Presentation</a> interface.
+        </p>
+        <pre class="idl">
+          interface Presentation {
+            // This API used by controlling browsing context.
```
mfoltzgoogle: ...API is used by...

index.html
```diff
@@ -1648,6 +1607,76 @@
...
+            attribute PresentationRequest? defaultRequest;
+
+            // This API is available on presenting browsing context.
+            [SameObject]
+            readonly attribute PresentationReceiver? receiver;
+          };
+
+</pre>
+        <p>
+          The <dfn for="Presentation"><code>defaultRequest</code></dfn> MUST
```
mfoltzgoogle: Is this requirement necessary?  I believe that this attribute works like any other null-able attribute, once set to a non-null value by the page it remains set.

index.html
```diff
@@ -1648,6 +1607,76 @@
...
+          return the <a>default presentation request</a> if any, <code>null</code>
+          otherwise.
```
mfoltzgoogle: On the other hand what is the effect of setting `defaultRequest` on a UA that implements the receiver but not the controller?  I assume that it's a no-op, i.e. the associated algorithms would not execute.

index.html
```diff
@@ -1648,6 +1607,76 @@
...
+        </p>
+        <p>
+          The <dfn for="Presentation"><code>receiver</code></dfn> MUST return
```
mfoltzgoogle: ...MUST be <code>null</code>...

as `receiver` is not a function.

mfoltzgoogle:
Looks good - with some editorial suggestions.

I agree with @tidoust that we should revisit the behavior of this API once the document is restructured around conformance classes, and nail down what everything does in each.

index.html

index.html
mounirlamouri: In the spirit of keeping commits focused on one thing, I will do that in a follow-up.

index.html
mounirlamouri: Added "the". Will change to "receiving browsing context" in follow-up.

index.html

index.html
mounirlamouri: I think it's good to have that. FWIW, I've only moved things around. I don't think I've added this.

index.html
mounirlamouri: Correct. The "Setting the default presentation request" section would allow that, IMO.

index.html
mounirlamouri: "return ```null```" is the correct terminology. As far as Web IDL is concerned, this is defining a getter for the attribute.

mounirlamouri:
Comments applied, merging. Thanks for the review!

PR #188, opened 2015-09-08
Address timeless' editorial feedback

anssiko:
A PR with some editorial changes on behalf of timeless. For the context, see: https://lists.w3.org/Archives/Public/public-secondscreen/2015Aug/0147.html


mounirlamouri:
lgtm, thanks

PR #191, opened 2015-09-12
Rename "presenting browsing context" to "receiving browsing context".

mounirlamouri:


index.html
```diff
@@ -450,7 +450,7 @@
       </section>
       <section>
         <h3>
-          Presenting context: monitor available session(s) and say hello.
+          Receiving context: monitor available session(s) and say hello.
```
mfoltzgoogle: At this point I don't believe "Receiving context" has been defined yet in the prose?

index.html
```diff
@@ -514,10 +514,10 @@
           <a>presentation session</a> via a <code>sessionconnect</code> event.
         </p>
         <p>
-          The <dfn>presenting browsing context</dfn> (or <dfn data-lt=
+          The <dfn>receiving browsing context</dfn> (or <dfn data-lt=
           "presentation context">presentation</dfn> for short) is the browsing
```
mfoltzgoogle: Do we need a definition of "presentation context" any more?

index.html
```diff
@@ -1479,9 +1478,9 @@
                 <li>Set the <a>presentation session identifier</a> of
                 <var>S</var> to <var>I</var>.
                 </li>
-                <li>Establish the connection between the controlling and
-                presenting browsing contexts using an implementation specific
-                mechanism.
+                <li>Establish the connection between the controlling and <a data-lt=
```
mfoltzgoogle: Maybe define this term above as "receiving browsing context|receiving browsing contexts", similar to the controlling case.

mfoltzgoogle:
LGTM with comments applied.  Thanks for doing this @mounirlamouri!

index.html

index.html
mounirlamouri: Doesn't seem like it. Removed.

index.html

mounirlamouri:
Review comments applied. Merging.

PR #196, opened 2015-09-17
Rename "presenting browsing context" to "receiving browsing context" on "uc-req.md" file.

hyunjunekim:
Latest renamed "presenting browsing context" to "receiving browsing context". So should change it which is file called "uc-req.md".


anssiko:
Thanks for the fix. If you identify other terminology that is not in sync please feel free to submit another PR.

PR #197, opened 2015-09-17
Fix up wrong codes about "EXAMPLE6"

hyunjunekim:
Fix up Grammar error about EXAMPLE6.

### DIFF
--- a/index.html
+++ b/index.html
@@ -475,7 +475,7 @@
       if (evt.data == "say hello")
         session.send("hello");
     }
-  });
+  };
   navigator.presentation.receiver.getSession().then(addSession);
   navigator.presentation.receiver.onsessionavailable = function(evt) {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -475,7 +475,7 @@
       if (evt.data == "say hello")
         session.send("hello");
     }
```
anssiko: Could you update the PR to add the missing semicolon (;) on L477 so we fix that in the same PR, thanks!

hyunjunekim:
@anssiko fixed it.

anssiko:
Thanks, merged.

In general, it is a good practice to test the syntax is valid when amending the examples section. We have evidence that spec examples get copied by web developers. To help ensure the syntax errors do to proliferate, I suggest we use some code analysis tool when updating the examples. There are various, for example the online service http://jshint.com/ and its CLI http://jshint.com/install/.

Issue #171, opened 2015-08-24
Rename PresentationSession to PresentationController


avayvod:
Based on https://lists.w3.org/Archives/Public/public-secondscreen/2015Aug/0027.html
Seems like renaming to PresentationController is the best way, where close() would close the presentation and an added disconnect() would detach the controller from the presentation.

mounirlamouri:
FWIW, @sicking suggested PresentationConnection in the thread.

avayvod:
As I understand, it depends on issue #35:
- if the api is close() to shut down presentation and disconnect() for disconnecting from it, PresentationController makes more sense (so one closes presentation via controller and disconnects the controller from it, PresentationConnection.close() would be confusing and disconnect() would sound redundant);
- if the api is close() for disconnecting and end()/stop() for actually terminating the presentation, PresentationConnection makes more sense.

Since we already have close() that means tearing down the presentation, disconnect() seems an easier path to achieve what we want, therefore PresentationController is my choice. Jonas didn't seem to have a strong opinion.

mfoltzgoogle:
I agree with @avayvod's point of view on this.  Let's close #35 and use the result to decide this.

sicking:
One issue is that this interface lives on both the presentation and the controller side. Something like `PresentationController` sounds to me like an object that only represents the controller side. In that regard `PresentationSession` or `PresentationConnection` is IMHO better.

I agree that it probably makes sense to make the final decision here once #35 is resolved.

mfoltzgoogle:
@avayvod given that we have `.close()` for disconnecting and `.terminate()` for stopping presentation, the argument here lands us at PresentationConnection.  Do you agree?

avayvod:
Yep. SGTM

mfoltzgoogle:
SGTM

Issue #152, opened 2015-07-28
Reconcile sessionavailable vs. sessionconnect for notifying page of session connection


mfoltzgoogle:
PR #146 adds a `sessionconnect` event to notify the controlling page that a presentation session is now accessible to that page.  However we previously introduced `sessionavailable` for this purpose on the presenting side.  I propose we use one event for both sides.

`sessionconnect` may not be the best name because on the controlling side, the event is fired before the session becomes connected.  Therefore I propose naming the event `sessionavailable` to denote that the session may be disconnected or connected when the event fires.

mfoltzgoogle:
We should resolve this after #171 is closed, as we may end up with `connectionavailable` instead.

mounirlamouri:
We just discussed this with @avayvod and we agreed that given the changes from #171 we should rename the event ```connectionavailable```.

@mfoltzgoogle if you do the changes for #171, could you apply that change too?

mfoltzgoogle:
connectionavailable SGTM

tidoust:
While we're renaming interfaces, would it also make sense to drop `session` from `PresentationSessionConnectEvent`, leading to `PresentationConnectEvent`?

mounirlamouri:
I believe this should be called ```PresentationConnectionAvailableEvent``` given the recent renamings agreed upon.

PR #203, opened 2015-10-06
Renames PresentationSession to PresentationConnection

mfoltzgoogle:
This addresses Issues #152  and #171 by renaming "presentation session" to "presentation connection," `PresentationSession` to `PresentationConnection`, and the events renamed as in #152. 

For the most part the changes are mechanical; however, in some places, the term "presentation connection" did not make sense and was replaced by "presentation."

Also, this PR has the side effect of redefining the presentation ID and URL as inherent to the presentation, not of an individual connection to the presentation. They continue to be provided as readonly properties on each `PresentationConnection.`  The algorithms may need some adjustment to ensure that matching and iteration based on URL and ID make sense.

### DIFF
--- a/index.html
+++ b/index.html
@@ -363,30 +363,30 @@
         <pre class="example highlight">
 &lt;!-- controller.html --&gt;
 &lt;script&gt;
-  // Start new connection.
+  // Start new presentation.
   request.start()
-    // the new started connection will be passed to setConnection on success
+    // The connection to the presentation will be passed to setConnection on success.
     .then(setConnection)
-    // user cancels the selection dialog or an error is occurred
+    // User canceled the selection dialog or an error occurred.
     .catch(endConnection);
 &lt;/script&gt;
 </pre>
       </section>
       <section>
         <h3>
-          Reconnect to a presentation connection example
+          Reconnect to a presentation example
         </h3>
         <pre class="example highlight">
 &lt;!-- controller.html --&gt;
 &lt;script&gt;
   // read presId from localStorage if exists
   var presId = localStorage &amp;&amp; localStorage["presId"] || null;
-  // presId is mandatory when reconnecting to a connection.
+  // presId is mandatory when reconnecting to a presentation.
   if (presId) {
     request.reconnect(presId)
-      // The reconnected connection will be passed to setConnection on success
+      // The resumed connection to the presentation will be passed to setConnection on success.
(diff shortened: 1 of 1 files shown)


tidoust:
That looks good to me. I note that @mounirlamouri had suggested an interface name closer to the event name for `PresentationConnectEvent`. I don't feel strongly one way or the other.

Not triggered by this pull request but, unless I missed something, there seems to be a bug in ReSpec that makes it link all occurrences of "presentation" to the IDL definition, despite the presence of a `data-lt` attribute that should rather make it point to the definition of "receiving browsing context". I created an issue on ReSpec for that:
https://github.com/w3c/respec/issues/510

index.html
```diff
@@ -1487,36 +1489,36 @@
...
           <h3>
-            Interface <code>PresentationSessionConnectEvent</code>
+            Interface <code>PresentationConnectEvent</code>
           </h3>
           <pre class="idl">
-            [Constructor(DOMString type, PresentationSessionConnectEventInit eventInitDict)]
-            interface PresentationSessionConnectEvent : Event {
-              [SameObject] readonly attribute PresentationSession session;
+            [Constructor(DOMString type, PresentationConnectEventInit eventInitDict)]
+            interface PresentationConnectEvent : Event {
```
mounirlamouri: This should be name PresentationConnectionAvailableEvent IMO

index.html
```diff
@@ -1487,36 +1489,36 @@
...
         <section>
           <h3>
-            Interface <code>PresentationSessionConnectEvent</code>
+            Interface <code>PresentationConnectEvent</code>
           </h3>
           <pre class="idl">
-            [Constructor(DOMString type, PresentationSessionConnectEventInit eventInitDict)]
-            interface PresentationSessionConnectEvent : Event {
-              [SameObject] readonly attribute PresentationSession session;
+            [Constructor(DOMString type, PresentationConnectEventInit eventInitDict)]
```
mounirlamouri: This should be name PresentationConnectionAvailableEventInit IMO.

mounirlamouri:
Thanks!

lgtm with the two name changes. Given the size of the diff, feel free to do that in a follow-up.

mfoltzgoogle:
Thanks for the review @tidoust and @mounirlamouri.  Merging PR with comments applied.

PR #204, opened 2015-10-08
Defines two conformance classes for the specification

mfoltzgoogle:
This PR defines two conformance classes for the specification to address Issue #93:

* A controlling user agent provides the controlling browsing context and its associated interfaces/objects.
* A receiving user agent provides the receiving browsing context and its associated interfaces/objects.

The PR makes normative statements about each of them where it makes sense to do so.  In most cases I say that a particular user agent SHOULD or MUST implement a certain interface in the corresponding browsing context - perhaps there is a better way to make these normative statements.

I didn't try to regroup the definitions by conformance class in this PR, but that might make the spec read better in the long run.  For example, PresentationConnection is required by both UAs.  It may be clearer to have a common definition and then the management algorithms in UA-specific sections. I think that is better done in a different PR.

@tidoust please take a look as you had some specific thoughts about how to describe the conformance classes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -800,8 +800,7 @@
             Terminating a presentation
           </h4>
           <p>
-            When a <a>controlling user agent</a> is to <dfn data-lt=
-            "terminate-algorithm-controller">terminate a presentation</dfn>
+            When a <a>controlling user agent</a> is to terminate a presentation
             using <em>connection</em>, it MUST run the following steps:
           </p>
           <ol>
@@ -832,17 +831,15 @@
                     </li>
                   </ol>
                 </li>
-                <li>Signal the <a>receiving user agent</a> to <a data-lt=
-                "terminate-algorithm-receiver">terminate the presentation</a>
-                using an implementation specific mechanism.
+                <li>Signal the <a>receiving user agent</a> to terminate the
+                presentation using an implementation specific mechanism.
                 </li>
               </ol>
             </li>
           </ol>
           <p>
-            When a <a>receiving user agent</a> is to <dfn data-lt=
-            "terminate-algorithm-receiver">terminate a presentation</dfn> using
-            <em>connection</em>, it MUST close the <a>receiving browsing
+            When a <a>receiving user agent</a> is to terminate a presentation
+            using <em>connection</em>, it MUST close the <a>receiving browsing
             context</a> (equivalent to calling <code>window.close()</code>).
           </p>
           <p>
@@ -1654,7 +1651,7 @@
                 <li>
                   <a>Queue a task</a> to <a>fire</a> an event named
                   <code>connectionavailable</code> at <a for=
-                  "Navigator"><code>presentation.receiver</code></a>.
+                  "PresentationReceiver"><code>presentation.receiver</code></a>.
                 </li>
               </ol>
             </li>
--- a/index.html
+++ b/index.html
@@ -203,9 +203,9 @@
         <a data-lt="receiving browsing context">presentation</a> run on the
         same UA and the operating system is used to route the <a>presentation
         display</a> output to the <a>presentation display</a>. This is commonly
-        referred to as the <b id="1-ua">1-UA</b> case. This specification
-        imposes no requirements on the <a>presentation display</a> devices
-        connected in such a manner.
+        referred to as the <dfn><b id="1-ua">1-UA</b></dfn> case. This
+        specification imposes no requirements on the <a>presentation
+        display</a> devices connected in such a manner.
       </p>
       <p>
         If the <a>presentation display</a> is able to render HTML documents and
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -247,6 +247,54 @@
...
+          <dd>
+            <p>
+              Web browsers that conform to the specifications of a
+              <a>controlling user agent</a> must be able to start and control
+              presentations by providing a <a>controlling browsing context</a>
+              as described in this specification. This context MUST implement
+              the <a><code>PresentationConnection</code></a>,
+              <a><code>PresentationConnectionAvailableEvent</code></a>, and
+              <a><code>PresentationRequest</code></a> interfaces.
+            </p>
```
tidoust: s/This context MUST implement/A controlling user agent implements/
In other words, it is the user agent that implements the interfaces and I would not include normative statements directly in the definition of the conformance class, but merely state facts.

The "Presentation" interface should be in that list, too.

index.html
```diff
@@ -247,6 +247,54 @@
...
+          <dd>
+            <p>
+              Web browsers that conform to the specifications of a <a>receiving
+              user agent</a> must be able to render presentations by providing
+              a <a>receiving browsing context</a> as described in this
+              specification. This context must implement the
+              <a><code>PresentationConnection</code></a>,
+              <a><code>PresentationConnectionAvailableEvent</code></a>, and
+              <a><code>PresentationReceiver</code></a> interfaces.
+            </p>
```
tidoust: Same comments as above: s/This context must implement/A receiving user agent implements/ and add "Presentation" interface as well.

index.html
```diff
@@ -247,6 +247,54 @@
...
+          </dd>
+        </dl>
+        <p>
+          One user agent may act both as an <a>controlling user agent</a> and
```
tidoust: s/an/a

index.html
```diff
@@ -247,6 +247,54 @@
...
+          as a <a>receiving user agent</a>, if it provides both browsing
+          contexts and implements all of their required interfaces. This can
+          happen when the the same user agent is able to host the
```
tidoust: s/the the/the

index.html
```diff
@@ -876,14 +927,21 @@
...
 </pre>
         <p>
           A <a><code>PresentationAvailability</code></a> object is associated
-          with a <a>presentation display</a> and represents its
-          <dfn>presentation display availability</dfn>.
+          with available <a>presentation displays</a> and represents the
+          <dfn>presentation display availability</dfn> for a presentation
+          request. The <a><code>PresentationAvailability</code></a> object
+          SHOULD be implemented in a <a>controlling browsing context</a> by a
+          <a>controlling user agent</a>, but may be omitted if the
```
tidoust: s/may/MAY. Actually, I wonder if that should not be "MUST be implemented... unless", since I don't see a way not to implement the interface if the user agent supports monitoring.

index.html
```diff
@@ -1709,53 +1793,47 @@
...
+            Controlling user agent
           </h4>
           <p>
+            The following partial interface MUST be implemented by a
+            <a>controlling user agent</a>:
+          </p>
+          <pre class="idl">
+            partial interface Presentation {
+              attribute PresentationRequest? defaultRequest;
+            };
```
tidoust: I believe that you can only use `partial` to extend an existing interface definition, but we don't have such a definition here. You would need to start with an empty "interface Presentation {};" somewhere.

There may be a way to avoid ending up with such an empty interface. We could stick to only one IDL block and explain in prose what each type of user agent needs to support. The advantage of using partial interfaces is that it clearly shows what will be available where.

index.html
```diff
@@ -1770,6 +1848,28 @@
...
         </section>
+        <section>
+          <h4>
+            Receiving user agent
+          </h4>
+          <p>
+            The following partial interface MUST be implemented by a
+            <a>receiving user agent</a>:
+          </p>
+          <pre>
```
tidoust: Missing `class="idl"`.
Hmm, trying locally (with an empty interface definition), I see that ReSpec is getting really confused about references to the Presentation interface :(

index.html
```diff
@@ -1770,6 +1848,28 @@
...
+            readonly attribute PresentationReceiver? receiver;
+          };
+          
+</pre>
+          <p>
+            The <dfn for="Presentation"><code>receiver</code></dfn> MUST return
+            <code>null</code> if the current <a>browsing context</a> is not a
+            <a>receiving browsing context</a> and return a
+            <a><code>PresentationReceiver</code></a> instance otherwise.
+          </p>
```
tidoust: Can the current browsing context not be a receiving browsing context for a receiving user agent?

In the end, keeping just one `Presentation` interface might make our lives easier (on top of other things, ReSpec seems to lose itself if we use two different partial definitions). The prose could simply say that:

1. `defaultRequest` will always return null in a receiving user agent that is not a controlling user agent as well.
2. `receiver` will always return null in a controlling user agent that is not a receiving user agent as well.
3. a controlling user agent MUST NOT implement the `PresentationReceiver` interface, even if it appears in the interface definition
4. a receiving user agent MUST NOT implement the `PresentationRequest` interface, even if it appears in the interface definition.

Note I say "must not" above because the existence of these interfaces should mean that they can be used!

What do you think?

index.html
```diff
@@ -551,7 +599,9 @@
         </h3>
         <p>
           Each <a>presentation connection</a> is represented by a
-          <a>PresentationConnection</a> object.
+          <a>PresentationConnection</a> object. Both the <a>controlling user
+          agent</a> and <a>receiving user agent</a> must implement
```
tidoust: s/must/MUST

tidoust:
@mfoltzgoogle,

I made a few comments inline, the main point for me being the definition of the `Presentation` interface in the spec. The interface is common to both user agents but specific for each of them. I think I initially suggested to use two `partial` definitions, but that looks a bit strange in practice for such a simple interface. In the end, I think I would stick to one `Presentation` interface definition, using regular prose to explain what each user agent implements.

This looks pretty good to me otherwise!

index.html

index.html
mfoltzgoogle: Done!

index.html

index.html
mfoltzgoogle: Looks okay to me...

mfoltzgoogle: NM done!

index.html
mfoltzgoogle: I reprhased it to be more normative as you suggest.  We may also want to make a normative statement for user agents that cannot implement background monitoring - either omit the object entirely, or implement it and always reject `getAvailability()`.

index.html
mfoltzgoogle: I started with that but the prose felt awkward, since there is clearly one part of the interface that applies only to a controlling UA and another that applies to a receiving UA.

index.html
mfoltzgoogle: Good catch, done.

index.html
mfoltzgoogle: > Can the current browsing context not be a receiving browsing context for a receiving user agent?

I believe the implication is that all browsing contexts on the receiving UA would have `navigator.presentation.receiver` defined but set to null if the context is not a presentation.  @mounirlamouri I believe you added this definition?

> defaultRequest will always return null in a receiving user agent that is not a controlling user agent as well.

What if script attempts to set this property - will it still return null?  That seems a little odd.
Perhaps we should hide the property instead?

> receiver will always return null in a controlling user agent that is not a receiving user agent as well.

Done.

Your proposal in #3 and #4 seems to prevent a single user agent from implementing both the controlling and  receiving browsing contexts!  This puts Chrome or any other 1-UA mode browser in a tough situation :-)

index.html

mfoltzgoogle:
@tidoust Thanks for all the comments.  I went ahead and merged the two interfaces and use prose to make statements about each property of the interface as you suggested.

We may want to also make statements to the effect that a non-controlling UA must not define `navigator.presentation.defaultRequest` and a non-receiving UA must not define `navigator.presentation.receiver`.  But I am not sure.

I am going to merge this, feel free to make additional suggestions/PRs.

PR #212, opened 2015-10-21
Editorial changes

anssiko:
Some editorial changes and clarifications that I spotted while reviewing the first half of the spec.


mfoltzgoogle:
LGTM, thanks @anssiko!

PR #213, opened 2015-10-22
Use concise, contemporary style for exceptions

anssiko:
* Update the spec to use the more concise style for exceptions that is adopted by contemporary specifications.
* Add definitions for the exceptions used by this spec to Terminology.
* Tidy.


mfoltzgoogle:
Thanks @anssiko!

Issue #201, opened 2015-10-06
PresentationReceiver: rename getConnection() and getConnections()


mounirlamouri:
CC @avayvod @mfoltzgoogle @sicking

I have looked at the receiver side of the API fairly rarely and every time I am to remember that getConnection() is actually waiting for a connection while getConnection***s*** return the list of active connections. I would suggest to rename them to:
```idl
interface PresentationReceiver {
  Promise<PresentationConnection> waitForConnection();
  Promise<sequence<PresentationConnection>> connections;
  attribute EventHandler onconnectionavailable;
};
```

I'm dubious about ```waitForConnection``` because it's basically the same as the ```connectionavailable``` event or did I miss the subtlety here?

mounirlamouri:
@avayvod and I talked to @jakearchibald and we came up with this proposition:
```idl
interface PresentationReceiver {
  Promise<> ready;
  sequence<PresentationConnection> connections;
  attribute EventHandler onconnectionavailable;
};
```

It made me wonder why we couldn't simply do:
```idl
interface PresentationReceiver {
  sequence<PresentationConnection> connections;
  attribute EventHandler onconnectionavailable;
};
```

It seems that if we accept that ```connections``` can be returned synchronously and the event will be fired when there is a new connection, pages could have code like:
```js
function handshake(c) {
   // ...
}

function onloadhandler() {
  navigator.presentation.receiver.onconnectionavailable = handshake;
  navigator.presentation.receiver.connections.forEach(handshake);
}
```

WDYT?

sicking:
I don't have a strong opinion either way. But if we do have a synchronous .connections property, we need to define when exactly the connection which launched the page on the TV shows up there. More specifically, is .connections.length==1 when the page on the TV is first launched?

@schien should probably comment on how much work it would be to make the initial connection synchronously available.

avayvod:
The F2F meeting notes about the issue 19 that led to the current spec: http://www.w3.org/2015/05/19-webscreens-minutes.html#item12

The argument against having an array property was that the prototype can be changed by the web developers. It doesn't apply to Sequence though afaik.
The other argument against the proposal was about supporting a simple use case of one connection - then the page wouldn't need an event handler or care about a sequence of the connections. This is why we left getSession() method.

@sicking I think the timing doesn't really matter in the Mounir's example - if the connection is already initialized, it could be populated by the time the page accesses the sequence but if it's not, then the page will receive an event.

[...]

Issue #193, opened 2015-09-14
Define a "set of presentations" for the receiving context?


tidoust:
The "set of presentations" is defined as "the presentation sessions created by the controlling browsing contexts for the user agent'. However, the `PresentationReceiver` interface also makes use of that set, e.g. in `getSession()`: "If the set of presentations is empty, then wait until at least one element is added to the set of presentations".

However, if the UA implements both the controller and receiver interfaces, the first presentation session added to this set may not be an "incoming" presentation session. It could be a session created through a call to `start()` instead.

The prose could perhaps be adjusted to distinguish entries in that set based on the URL (null for incoming presentations, set for those created). However, it seems easier to define a separate "set of presentations" for the receiving context, which could perhaps be named "set of incoming presentations".

What do you think?

mfoltzgoogle:
We need to distinguish the following three sets, as they are distinct:

- The set of presentation sessions created for a controlling browsing context via start/reconnect
- The set of presentation browsing contexts acting as a presentation receivers
- The set of presentation sessions provided to the presentation receiver

mfoltzgoogle:
Also from PR #221:

> The text about "PresentationReceiver object" should probably move
> to the section about the Presentation interface. This section is
> about the interface, not its instance (the object).

> Furthermore, we probably need a better way to explain how
> PresentationConnectionList is initialized -- how the set of
> presentations is populated.

PR #221, opened 2015-11-19
Update PresentationReceiver spec to match the results from the discussions in #201

mounirlamouri:


mounirlamouri:
/CC @sicking @avayvod @mfoltzgoogle 

One difference between this PR and the discussion we had is that `.connections` return an array instead of a sequence<> the reason being that we can't return a sequence<> from an attribute.

sicking:
We should return a FrozenArray<T>. That's the best way to have an attribute which returns an list of things.

mounirlamouri:
Thanks for the info Jonas. I switched to that :)

index.html
```diff
@@ -1736,24 +1718,20 @@
       </section>
       <section>
         <h3>
-          Interface <a><code>PresentationReceiver</code></a>
+          Interface <a><code>PresentationConnectionList</code></a>
```
mfoltzgoogle: Are we using this for both the controlling and receiving browsing contexts?

We could keep this as a generic interface that could be implemented by either or both, but if we need to add additional features that are specific to the receiving context I feel we would end up defining a PresentationReceiver interface again anyway.

index.html
```diff
@@ -1736,24 +1718,20 @@
...
 </pre>
         <p>
-          The <a>PresentationReceiver</a> object is available to a <a>receiving
-          browsing context</a> in order to access the <a data-lt=
-          "controlling browsing context">controlling browsing context</a> and
-          communicate with it. The <a><code>PresentationReceiver</code></a>
-          object MUST be implemented in a <a>receiving browsing context</a>
-          provided by a <a>receiving user agent</a>.
+          The <a for="PresentationConnectionList">connections</a> attribute
+          MUST return a the non-terminated <a>set of presentations</a>
```
mfoltzgoogle: Typo in "a the"

mfoltzgoogle: Does non-terminated mean connection.state != terminated?  How about closed?

index.html
```diff
@@ -1797,23 +1775,22 @@
                 </li>
                 <li>
                   <a>Queue a task</a> to <a>fire</a> an event named
-                  <code>connectionavailable</code> at <a data-lt=
-                  "PresentationReceiver"><code>presentation.receiver</code></a>.
+                  <code>connectionavailable</code> at all
```
mfoltzgoogle: Can there be more than one list per receiver?

index.html
```diff
@@ -1797,23 +1775,22 @@
...
-            when a connection associated with the object is created. It is
-            fired at the <a>PresentationReceiver</a> instance, using the
-            <a>PresentationConnectionAvailableEvent</a> interface, with the
-            <a for="PresentationConnectionAvailableEvent">connection</a>
-            attribute set to the <a><code>PresentationConnection</code></a>
-            object that was created. The event is fired for all connections
-            that are created for the <a>presentation</a> when <a>monitoring
-            incoming presentation connections</a>.
+            <code>connectionavailable</code> on all
+            <a>PresentationConnectionList</a> when a connection is created. It
```
mfoltzgoogle: Similar comment about firing the event at a single intended PresentationConnectionList.

index.html
```diff
@@ -1797,23 +1775,22 @@
...
+            is fired using the <a>PresentationConnectionAvailableEvent</a>
+            interface, with the <a for=
+            "PresentationConnectionAvailableEvent">connection</a> attribute set
+            to the <a><code>PresentationConnection</code></a> object that was
+            created. The event is fired for all connections that are created
```
mfoltzgoogle: I might say "for each connection that is created"

index.html
```diff
@@ -1934,6 +1834,66 @@
...
+          communicate with it. The <a><code>PresentationReceiver</code></a>
+          object MUST be implemented in a <a>receiving browsing context</a>
+          provided by a <a>receiving user agent</a>.
+        </p>
+        <p>
+          The <a for="PresentationReceiver">connections</a> attribute MUST
+          return the result of running these steps:
+        </p>
+        <ol>
+          <li>Let <var>p</var> be a new <a>Promise</a>.
```
mfoltzgoogle: Does this need to return a new Promise each time it's referenced in script?  Or should the algorithm return the same Promise each time?

index.html
```diff
@@ -1934,6 +1834,66 @@
...
+          <li>Run the following substeps asynchronously:
+            <ol>
+              <li>Let <var>list</var> be a new
+              <a>PresentationConnectionList</a>.
+              </li>
+              <li>Initialize <var>list</var> with the <a>set of
+              presentations</a> associated with the <a>receiving browsing
+              context</a>.
+              </li>
+              <li>If the user agent is not <a>monitoring incoming Presentation
```
mfoltzgoogle: nit: s/Presentation/presentation/

index.html
```diff
@@ -1934,6 +1834,66 @@
...
+        <div class='issue'>
+          <p>
+            The text about "PresentationReceiver object" should probably move
+            to the section about the Presentation interface. This section is
+            about the interface, not its instance (the object).
+          </p>
+          <p>
+            Furthermore, we probably need a better way to explain how
+            PresentationConnectionList is initialized -- how the set of
+            presentations is populated.
```
mfoltzgoogle: The place to do this is in 6.4.6, "Establishing a presentation connection in a receiving browsing context" which starts monitoring for incoming presentation connections.  The set of presentations for the receiving context would be initialized there I think.

The "set of presentations" data structure in general needs some cleanup, feel free to drop a reference to https://github.com/w3c/presentation-api/issues/193 instead here.

mfoltzgoogle:
Sorry for delays, I blame the holidays :)
Mostly looks good with a few comments inline.

index.html
mounirlamouri: Ack :)

index.html
mounirlamouri: Typo done.

"non-terminated" indeed means state being not terminated. When we discussed this in the issue, we decided to include "closed" connections because they might re-open.

index.html
mounirlamouri: The idea is that there is only one list per receiver, yes. Is this a misconception?

index.html
mounirlamouri: Done

index.html
mounirlamouri: Good point. Updated.

index.html

index.html
mounirlamouri: Not entirely sure what you wanted me to do here. Added a link to issue 193.

mounirlamouri:
PTAL :)

index.html
mfoltzgoogle: So the step to "fire an event named connectionavailable at all PresentationConnectionList" should only target the objects that are associated with the receiving browsing context in this instance of the algorithm, not every object in the receiving user agent.   Just saying "...at the PresentationConnectionList associated with the receiving browsing context" would be fine for now.

index.html
```diff
@@ -1934,6 +1834,74 @@
...
+          Interface <a><code>PresentationReceiver</code></a>
+        </h3>
+        <pre class="idl">
+          interface PresentationReceiver {
+            [SameObject] readonly attribute Promise&lt;PresentationConnectionList&gt; connections;
+          };
+
+
+</pre>
+        <div class='issue'>
```
mfoltzgoogle: Nit: use data-number="193" to auto-include a link to GH.

index.html
```diff
@@ -1934,6 +1834,74 @@
...
+          <a>PresentationConnectionList</a>.
+        </p>
+        <ol>
+          <li>If <a>connection list promise</a> is not <code>null</code>,
+          return <a>connection list promise</a> and abort these steps.
+          </li>
+          <li>Otherwise, set <a>connection list promise</a> be a new
+          <a>Promise</a>.
+          </li>
+          <li>Return <var>p</var>.
```
mfoltzgoogle: Return <a>connection list promise</a>

index.html
```diff
@@ -1934,6 +1834,74 @@
...
+              <li>Initialize <var>list</var> with the <a>set of
+              presentations</a> associated with the <a>receiving browsing
+              context</a>.
+              </li>
+              <li>If the user agent is not <a>monitoring incoming presentation
+              connections</a>, <a>queue a task</a> to start <a>monitoring
+              incoming presentation connections</a> from <a>controlling
+              browsing contexts</a>.
+              </li>
+              <li>Resolve <var>p</var> with <var>list</var>.
```
mfoltzgoogle: Resolve connection list promise

mfoltzgoogle:
I'll merge now and make these fixes separately. Thanks @mounirlamouri

Issue #216, opened 2015-10-29
Spec should clarify the initial presentation connection state


tidoust:
Example 5 in the spec (as discussed in #198) currently assumes that the initial state of the connection is `closed`, and that it will change to `open` afterwards, triggering a `statechange` event.

This matches my reading of the algorithms defined in the spec (which "queue a task" at key points to establish the communication channel and return the presentation in the meantime).

However, discussions at the F2F suggest that may not be the case.

The specification should be clear about the initial state of the presentation object returned when the start promise is resolved, or on the lack of guarantee that it has any initial state (for instance, because the Web app may check the resolved promise at any time and not necessarily as soon as the promise is resolved internally).

Developers need that clarification to know whether they can expect a `statechange` event to being with or whether they have to listen to the `statechange` event *and* check the initial state, if they want to do something with the presentation object when the channel is opened.

mfoltzgoogle:
I believe PR #227 (and the updated example) addresses this, please let me know if there are further clarifications needed.

Issue #214, opened 2015-10-23
Add a "starting" state to PresentationConnection


mfoltzgoogle:
When calling PresentationRequest.start(), it may take some time for the presentation to be fully loaded and ready for connection on the presentation screen.

In the meantime, the controlling page is unaware of the initialization progress of the presentation, since we only resolve the Promise with a connected PresentationConnection.  Activities (i.e., media playback) will continue in the meantime, while the user has already indicated they want the content to be presented remotely, which creates a jarring user experience.   If the controlling page were aware that presentation was starting, it could pause the media and update the UI to show the user what was happening.

@avayvod has proposed adding a starting state to the PresentationConnection, with a new start algorithm that would roughly go as follows:

1. If the user selects a screen, then resolve the Promise right away with a "starting" PresentationConnection. 
2. If the user cancels presentation, reject the Promise.
3. Once the receiving context is fully loaded, transition the PresentationConnection to connected.
4. If there is an error loading the receiving context then transition to disconnected (or perhaps a new error state).

Messaging would be allowed during the "starting" state in case the presentation wants to convey metadata about the chosen screen, or show some progress information.

mounirlamouri:
This is somethnig we wanted to add to the Remote Playback API too. I think it would be great to have this state. Though, we might want to call it `connecting` for consistency with `connected`?

avayvod:
Since sending and receiving messages to/from a "connecting" connection might sound confusing, and w/o this extra functionality "connecting" state doesn't add much, maybe there're other ways we could workaround passing the metadata etc?
1. Return the PresentationConnection in a "disconnected" state when the user picks a screen, but with some custom dictionary "screen" field that would contain the available metadata? Once the connection is established, the PresentationConnection changes to a "connected" state.
2. Return the PresentationConnection in a "connected" state when the user picks a screen to allow messaging and handle the messages requiring a working connection by queueing them until the underlying connection to the screen is established.

tidoust:
Hi @avayvod,

> Since sending and receiving messages to/from a "connecting" connection might sound confusing, and w/o this extra functionality "connecting" state doesn't add much, maybe there're other ways we could workaround passing the metadata etc?

While there may be better ways to expose messaging channels, consistency across similar specs seems like a good idea.

Typically, both WebSockets and WebRTC data channels have `connecting`, `open`, `closing` and `closed` ready states:
https://html.spec.whatwg.org/multipage/comms.html#dom-websocket-readystate
[...]

Issue #198, opened 2015-09-18
Message sending in Example 5 happens too early


tidoust:
The code in example 5 apparently assumes that it may issue a `send` command right away:
http://w3c.github.io/presentation-api/#monitor-session-s-state-and-exchange-data-example

However, I don't see what guarantees that this won't thrown an error in the spec. The algorithm to etablish a presentation connection runs in the background in particular, so I would expect the returned PresentationSession to be in "disconnected" state to start with:
http://w3c.github.io/presentation-api/#dfn-establish-a-presentation-connection

And the algorithm to send a message throws an exception if the state is not "connected":
http://w3c.github.io/presentation-api/#dfn-send-a-message-through-a-presentationsession-s

Happy to prepare a PR if I'm right.

louaybassbouss:
Hi @tidoust yes we should check the session's state before sending the message or we can put the send call in a `try-catch` block.

anssiko:
Reopened per F2F discussion.

tidoust:
See relevant discussion at TPAC F2F:
http://www.w3.org/2015/10/28-webscreens-minutes.html#item04

**ACTION**: Mark and Louay to rework example 5 and refine spec prose to clarify the initial connection state when the connection is created

mfoltzgoogle:
PR #227 modifies the example code to send messages only on a "connected" presentation connection.
It also creates new connections in a "connecting" state.

Issue #149, opened 2015-07-22
Define behavior when sending a message to a PresentationSession fails


mfoltzgoogle:
The spec does not define the outcome when the UA is unable to send a message to a connected PresentationSession.  As the spec includes a recommendation that the connection between sessions be treated as a reliable channel, it seems that we should notify the application when that guarantee cannot not be met.

Here are a few options (some are not mutually exclusive):

1.  Add an `error` event to the PresentationSession to notify the application that there is a failure to deliver the message.
2. Close the PresentationSession (which fires the `close` event).
3. Add a return value to (or return a Promise from) `send()` to convey the outcome of the message delivery.
4. Queue the message and attempt redelivery in case the underlying channel becomes reconnected at a future time.

I am thinking a combination of 1 (to let the application know a message was not delivered in a timely manner) and 4 (so that transient failures of the channel don't require intervention by the page).

We may want to allow the page to also set a timeout for message delivery that would result in outcome 2.

schien:
Agree with the proposal of combining 1 and 4. I remember this is the consensus during the previous f2f meeting.

obeletski:
Webrtc is using combination of 1 and 2 in case of sending failure [1], since they were considering case when output buffer is full and nothing can be done. I agree that we should wait for reconnect to happen if that is possible, so it would be 1 + 4 + after timeout 2.

1. http://w3c.github.io/webrtc-pc/#dom-datachannel-send

mounirlamouri:
I would be in favour of not automatically close and let the web page make that decision.

Regarding promise vs. event, I would prefer using a promise because it makes it clearer the error is about the message sending and it allows different error handling depending on the ```send()``` call. It also prevents us from having to pass back the message that was sent and all the issues with the type of the message. It comes with a bit `more overhead than the event though.

```js
session.onsenderror = function(e) {
  // Note: e.message? e.data? what should we expect the type to be?
  console.error("Was not able to send message '" + e.message + "'");
};
session.send(msg);
```
vs
```js
session.send(msg).catch(function(e) {
  console.error("Was not able to send message '" + msg + "'");  
});
```

If we go with the event, at least, I think we should not use ```error``` but ```senderror``` for the event name.

schien:
[...]

PR #227, opened 2015-12-15
Adds a new "connecting" state and per-state event handlers.

mfoltzgoogle:
This PR addresses the following issues:

* https://github.com/w3c/presentation-api/issues/214
* https://github.com/w3c/presentation-api/issues/149
* https://github.com/w3c/presentation-api/issues/198
* https://github.com/w3c/presentation-api/issues/216

Summary of changes:

* Adds a new "connecting" state to the PresentationConnection (Issue #214)
* Adds per-state event handlers `onconnected`, `onclosed` and `onterminated` to replace `onstatechange`
* Adds a `PresentationConnectionCloseEvent` and `PresentationConnectionCloseReason` with details on why a connection was closed, including an error in the underlying message transport (Issue #149).
* Refine example code and algorithms to clarify initial presentation states and ensure message sending happens at the right time.  (Issue #198, Issue #216)

### DIFF
--- a/index.html
+++ b/index.html
@@ -77,7 +77,7 @@
           }
         },
         issueBase: "https://www.github.com/w3c/presentation-api/issues/",
-        githubAPI: "https://api.github.com/repos/w3c/presentation-api",
+        // githubAPI: "https://api.github.com/repos/w3c/presentation-api",
 
         // Temp fix for ReSpec issue #483:
         // https://github.com/w3c/respec/issues/483
@@ -865,21 +865,6 @@
             <li>Otherwise, <em>T</em> has completed with the user <em>granting
             permission</em> to use a display; let <em>D</em> be that display.
             </li>
-            <li>
-              <a>Queue a task</a> <em>C</em> to run the following steps:
-              <ol>
-                <li>
-                  <a>Create a receiving browsing context</a> on <em>D</em> and
-                  let <em>R</em> be the result.
-                </li>
-                <li>
-                  <a>Navigate</a> <em>R</em> to <code>presentationUrl</code>.
-                </li>
-              </ol>
-            </li>
-            <li>If <em>C</em> fails, <a>reject</a> P with an
-            <a>OperationError</a> exception and abort all remaining steps.
-            </li>
             <li>Let <var>I</var> be a new <a>valid presentation identifier</a>
             unique among all <a>presentation identifiers</a> for known
             presentations in the <a>set of presentations</a>.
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
This is ready for review.  @sicking, @mounirlamouri, @avayvod and others, please take a look.

index.html
```diff
@@ -1497,37 +1624,60 @@
...
+            <dt>
+              Input
+            </dt>
+            <dd>
+              <code>presentationConnection</code>, the <a>presentation
+              connection</a> to be closed
+            </dd>
+            <dd>
+              <code>closeReason</code>, the
+              <code>PresenentationConnectionClosedReason</code> describing why
```
tidoust: Typo: "PresenentationConnectionClosedReason" should be "PresentationConnectionClosedReason"

avayvod:
The description mentions issue numbers inconsistently:
- the list has issues #217, #216, #196, #149 while the summary below only references #216, #196, #149 (so no #217) and a new #214;
- #149 doesn't seem to be related to per-state event handlers as the summary suggests though I might be wrong

index.html
```diff
@@ -1403,23 +1415,36 @@
...
+              <code>messageOrData</code>, the <a>presentation message data</a>
+              to send to the other browsing context
+            </dd>
+          </dl>
           <ol link-for="PresentationConnection">
-            <li>If the <a>state</a> property of <a>PresentationConnection</a>
-            is not <code>"connected"</code>, throw an
-            <code>InvalidStateError</code> exception.
+            <li>If the <a>state</a> property of
+            <code>presentationConnection</code> is not <code>connected</code>,
```
avayvod: In #214 we speculate about allowing messaging to/from PresentationConnection in a "connecting" state. Do you plan to add this in a follow up PR? Without this difference from "disconnected" state, it's not clear why we just not resolve the start() promise with the "disconnected" state that changes to "connected" later on. All I can think of is it will be clearer for the Web developer whether to expect the transition to "connected" or not but it might not be good enough to justify adding a new state then.

index.html
```diff
@@ -1449,11 +1500,31 @@
...
+            <dd>
+              <code>messageData</code>, the <a>presentation message data</a> of
+              the message
+            </dd>
+          </dl>
           <ol link-for="PresentationConnection">
-            <li>If the <a>state</a> property of <a>PresentationConnection</a>
-            is not <code>"connected"</code>, abort these steps.
+            <li>If the <a>state</a> property of
+            <code>presentationConnection</code> is not <code>connected</code>,
```
avayvod: Ditto.

index.html
mounirlamouri: I think an event based approach would be confused with a life cycle like `disconnected` -> `connected` -> `closed` -> `disconnected` because it could do:
```js
myConnecton.onstatechange = function() {
  switch (this.state) {
    case 'connected':
      showRemoteControlUI();
      break;
    case 'closed':
      addButtonToJoin();
      break;
    case 'disconnected':
      hideRemoteControlUI();
      break;
  }
};
```

mfoltzgoogle:
@avayvod Re: https://github.com/w3c/presentation-api/pull/227#issuecomment-165107424

Sorry about the poorly formatted description.  I've updated it to fix the issues you pointed out.

index.html
mfoltzgoogle: I think naming the state "connecting" is clearer than "disconnected" per consistency with other Web APIs (i.e. WebSockets)  and the example code above, as well as in the PR.

mfoltzgoogle:
Unless there's further feedback, I'll do one more editorial pass to clean up anything minor with this PR (i.e. fixing TODOs) then plan on merging.  As followups the following issues should be addressed:

* #228 - GitHub API failing, which leads to ReSpec failures
* #229 - Need algorithm for establishing a presentation connection when the previous connection(s) were closed
* #231 - Wrong use of "Queue a task" and "in parallel" in algorithms

Issue #231, opened 2015-12-17
Wrong use of "Queue a task" and "in parallel" in algorithms


tidoust:
I believe that some of the algorithms in the spec use "Queue a task" incorrectly.

"Queue a task" is strongly tied to the event loop:
http://www.w3.org/TR/html5/webappapis.html#queue-a-task
[[ When a user agent is to queue a task, it must add the given task to one of the task queues of the relevant event loop ]]

In particular, a task queued on the event loop does not run in the background and actually blocks the event loop and thus user interactions. However, there are several places in the specification where "Queue a task" seems used to mean something like "run the task in the background".

For example, in the `start` algorithm:
http://w3c.github.io/presentation-api/#dom-presentationrequest-start

[[ 9. Queue a task C to run the following steps:
  1. Create a receiving browsing context on D and let R be the result. 
  2. Navigate R to presentationUrl. ]]

Steps 9.1 and 9.2 do not have anything to do with the event loop of the current browsing context. Actually, there should be no need to use any particular keyword to introduce these steps: that part of the algorithm runs in the background already (since the algorithm returned the Promise to the calling script at step 4).

Same thing for substeps of step 3. in the `reconnect` algorithm that do not have anything to do with the event loop either, I think:
http://w3c.github.io/presentation-api/#dom-presentationrequest-reconnect
[[ 2. Return P.
3. Queue a task T to run the following steps in order: [...] ]]

This algorithm also returns at step 2, so remaining steps de facto run in the background. In fact, it would be a good idea to say so explicitly whenever we have a "Return" statement in the middle of an algorithm by saying "Return P and run the following steps in parallel" as done in other specs.

Which brings me to my second point: the `start` algorithm uses "in parallel" to mean "run the two following steps in two separate background processes", but "in parallel" is now used in other specs to mean:
https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel
[[ Some algorithms run in parallel; this means that the algorithm's subsequent steps are to be run, **one after another**, at the same time as other logic in the specification ]]

Step 5. in the `start` algorithm looks weird as a result, because I understand we want to run 5.1 and 5.2 as separate background tasks, while the definition suggests that they will be run in the background but in order:
[[5. Run the following steps in parallel:
  1. Monitor the list of available presentation displays. 
  2. Queue a task T to request user permission for the use of a presentation display and selection of one presentation display.]]

(There's another misuse of "Queue a task" here, as requesting user permission is also not tied to the event loop).

[...]

PR #239, opened 2016-01-12
Updated version of PR #238

mfoltzgoogle:
This is based on PR #238 but somewhat simplifies the two parts of the close algorithm and how they are invoked.  The first part signals the destination browsing context that the connection is closing and passes the reason (close or wentaway). The second part queues a task to change the connection state and fire the close event.

I wonder if we should change state before queueing a task on the event loop to avoid re-entrancy, to ensure correct behavior when calling close() then close() again (or another API method that checks the PresentationConnection state).

### DIFF
--- a/index.html
+++ b/index.html
@@ -1676,8 +1676,7 @@
           <p>
             When the <a>user agent</a> is to <dfn data-lt=
             "start closing the presentation connection|closing procedure">start
-            closing a presentation connection</dfn>, it MUST run the following
-            steps:
+            closing a presentation connection</dfn>, it MUST do the following:
           </p>
           <dl>
             <dt>
@@ -1746,17 +1745,17 @@
           <ol>
             <li>
               If there is a pending <a>close the presentation connection</a>
-              task, or the <a>close the presentation connection</a>
-              task has already run
-              for <code>presentationConnection</code>, then abort the
-              remaining steps.
+              task for <code>presentationConnection</code>, or a <a>close the
+              presentation connection</a> task has already run
+              for <code>presentationConnection</code>, then abort the remaining
+              steps.
             </li>
             <li>
               <a>Queue a task</a> to run the following steps:
               <ol>
                 <li>If the <a>presentation connection state</a> of
                 <code>presentationConnection</code> is not
-                <code>connecting</code>, <code>connected</code>
+                <code>connecting</code>, <code>connected</code>,
                 or <code>closed</code>, then abort the remaining steps.
                 </li>
                 <li>If the <a>presentation connection state</a>
--- a/index.html
+++ b/index.html
@@ -1698,14 +1698,26 @@
             </dd>
           </dl>
           <ol>
-            <li>Signal the intention to close the corresponding
-            <code>PresentationConnection</code> to the <a>destination browsing
-            context</a>, passing the <code>closeReason</code> to that context.
+            <li>If the <a>presentation connection state</a> of
+            <code>presentationConnection</code> not <code>connecting</code> or
+            <code>connected</code> then abort the remaining steps.
+            </li>
+            <li>If there is a pending task to <a>close the presentation
+            connection</a> <code>presentationConnection</code>, then abort the
+            remaining steps.
+            </li>
+            <li>Start to signal to the <a>destination browsing context</a> the
+            intention to close the corresponding
+            <code>PresentationConnection</code>, passing the
+            <code>closeReason</code> to that context. The user agent does not
+            need to wait for acknowledgement that the corresponding
+            <code>PresentationConnection</code> was actually closed before
+            proceeding to the next step.
             </li>
             <li>If <code>closeReason</code> is not <code>wentaway</code>, then
             locally run the steps to <a>close the presentation connection</a>
-            with <code>presentationConnection</code>,
-            <code>closeReason</code>, and <code>closeMessage</code>.
+            with <code>presentationConnection</code>, <code>closeReason</code>,
+            and <code>closeMessage</code>.
             </li>
           </ol>
           <p>
@@ -1732,13 +1744,14 @@
             </dd>
           </dl>
           <ol>
-            <li>If the <a>presentation connection state</a> of
-            <code>presentationConnection</code> is not <code>connecting</code>
-            or <code>connected</code>, then abort the remaining steps.
-            </li>
             <li>
               <a>Queue a task</a> to run the following steps:
               <ol>
+                <li>If the <a>presentation connection state</a> of
+                <code>presentationConnection</code> is not
+                <code>connecting</code> or <code>connected</code>, then abort
+                the remaining steps.
+                </li>
                 <li>Set the <a>presentation connection state</a> to
                 <code>closed</code>.
                 </li>
--- a/index.html
+++ b/index.html
@@ -1357,9 +1357,9 @@
             When the <code><dfn>close</dfn>()</code> method is called on a
             <a>PresentationConnection</a> <var>S</var>, the <a>user agent</a>
             MUST <a>start closing the presentation connection</a> <var>S</var>
-            with <code>closed</code> as <code>closeReason</code> and an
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
This is a followup to issue #231 (Wrong use of "Queue a task" and "in parallel" in algorithms).

index.html
```diff
@@ -1343,42 +1355,50 @@
...
             When the <code><dfn>close</dfn>()</code> method is called on a
-            <a>PresentationConnection</a>, the user agent MUST run the
-            algorithm to <a data-lt="close-algorithm">close a presentation
-            connection</a> with <a>PresentationConnection</a>, a
-            <code>reason</code> of <code>closed</code>, and an empty
-            <code>message</code>.
+            <a>PresentationConnection</a> <var>S</var>, the <a>user agent</a>
+            MUST <a>start closing the presentation connection</a> <var>S</var>
+            with <code>closed</code> as <code>closeReason</code> and an empty
+            <code>message</code> as <code>closeMessage</code>.
```
tidoust: The problems that I was trying to prevent in my PR here are:

1. ensure that two consecutive calls to close() won't "start closing the presentation connection" twice, to avoid signaling the other end twice and triggering two "closed" events.
2. ensure that calling close() on a presentation connection on which the closing procedure has already started through other means (object being discarded) would do nothing
3. ensure that the presentation connection state gets switched to "closed" immediately when close() is called, in other words I would expect the code in the `if` block below not to run, ever:
```
presentation.close();
if (presentation.state === "connected") {
  // This should never happen, presentation.state should already be "closed"
}
```

For 3., this would be consistent with WebSockets and RTCDataChannel (well, they use an intermediary "closing" state, but the point is that the presentation should no longer be "connected"). However, here, the "start closing the presentation connection" calls on the "close the presentation connection" procedure that "queues a task" to set the state to "closed". That task will run on next tick by definition. There needs to be some step here or in the closing procedure that instructs the user agent to set the state to "closed" immediately in the close() case.

I'm not entirely sure that 2. is possible, as the only other way to "start closing the presentation connection" is when 
the object is discarded, which might only happen at the very end when there are no longer any references to it. However, it seems to me that this could happen while there is still some JS code to run that holds a reference to the connection.

index.html
```diff
@@ -1688,40 +1698,62 @@
             </dd>
           </dl>
           <ol>
-            <li>If the <a>presentation connection state</a> of
-            <code>presentationConnection</code> is not <code>connected</code>
-            or <code>connecting</code>, then abort these steps.
+            <li>Signal the intention to close the corresponding
+            <code>PresentationConnection</code> to the <a>destination browsing
+            context</a>, passing the <code>closeReason</code> to that context.
             </li>
```
tidoust: One thing that does not strike me as clear is whether "signal" involves going through the closing handshake and thus waiting for the ack from the other side. If so, this should take place behind the scenes and I would suggest using "Start to signal", or clarifying what "signal" means somehow.

In particular, I assume that the "close the presentation connection" procedure is to run in the same run of the event loop when that procedure is called by the close() method.

index.html
```diff
@@ -1688,40 +1698,62 @@
...
+            </dd>
+            <dd>
+              <code>closeMessage</code>, a human-readable message with details
+              of why the connection was closed.
+            </dd>
+          </dl>
+          <ol>
+            <li>If the <a>presentation connection state</a> of
+            <code>presentationConnection</code> is not <code>connecting</code>
+            or <code>connected</code>, then abort the remaining steps.
```
tidoust: This step may be done there to avoid queueing a task for nothing, but it should also be done once the task runs to avoid firing the "closed" event twice, for instance if the procedure is run upon receipt of a signal from the other end (which happens in the background), while the JS code is calling close() (which happens in the event loop).

I would simply move it below as first step of the task. The possibility to check the state before queueing a task seems like an implementation detail.

index.html
mfoltzgoogle: I agree that cases 1 and 3 need to be addressed.  I will fix the PR to ensure that the connection state is set to 'closed' on entry to either the "starting" or "closing" procedures which should fix the re-entrancy cases.

I don't see a scenario where 2 should be possible (if the object is discarded by definition there should be no live references to it).

index.html
mfoltzgoogle: I'll try to clarify that the event loop should not block on receiving an ack from the destination user agent.  

Yes the steps in "start closing the presentation connection" should run in the same event loop when close() is called.   This seems to be implicit in this spec, i.e. it's assumed to be the case if we don't specify that we are queuing a task to create a new event loop.

index.html
mfoltzgoogle: Good point, will do.

mfoltzgoogle:
I've updated the PR which I think addresses the issues noted.

The connection is transitioned to `closed` in a task queued from the "close a presentation connection" steps.  Other attempts to close are no-ops if this task is pending, or the connection has already been closed.

I think this also prevents duplicate signals being sent to the destination browsing context, but please double check my logic throughout :)

index.html
tidoust: I think case 3 remains in your latest update: since the connection is transitioned to `closed` in a *task*, the transition won't happen in the same run of the event loop when `close()` is called, and thus the code in the "should never happen" block in my code example above will be executed.

We can't simply drop "queue a task" from the "close a presentation connection" procedure since that procedure is also called from other procedures that run in the background. Adding a step in this portion of the spec to switch the state to `closed` would mostly solve the issue, although the check on the `state` in the close procedure would need to be revisited as well :-(

mfoltzgoogle:
Okay, another attempt uploaded.

- It sets the connection state to `closed` immediately when `close()` is invoked, so the code from example 3 should be addressed.
- It makes the "close a connection" task run at most once per connection, the first thing that causes the connection to close "wins" and gets to set the reason and message, later procedures are aborted.

This would be simplified by adding a `closing` state to the connection which could be checked explicitly to prevent multiple close tasks from executing.  However this state would be used primarily to support the specification and it's not clear if it needs to be exposed to the API client.  Maybe add a side note that this internal state would be helpful to implementation?

tidoust:
This looks good to me!

I raised issue #240 the other day to track the `closing` state. I agree that the state seems only useful internally, but I may have missed something. I propose to add the idea of a side note to that discussion.

tidoust:
Based on the discussion in https://github.com/w3c/presentation-api/issues/240#issuecomment-171928315
... I think the `closed` event now fires too early. It should signal that there cannot be any incoming (or outgoing of course) message anymore on the connection, but communication protocols usually expect clients to continue listening to messages as long as the closing handshake is not over, which means that messages may still be received after a call to `close()` for a short amount of time.

My initial PR #238 was phrased along these lines, introducing the notion of a `closed` underlying connection, which triggered the close procedure, and saying that the closing procedure should eventually render the underlying connection closed:

```
<p>
  When the underlying connection of a <a>PresentationConnection</a>
  object <var>S</var> has been <dfn dfn-for=  
  "PresentationConnection">closed</dfn>, the <a>user agent</a> MUST  
  <a>close the presentation connection</a> <var>S</var> with the  
  reason and message of the initiating procedure, if any, or with  
  <code>closed</code> as <code>closeReason</code>, and an empty  
  <code>closeMessage</code> otherwise.  
</p>  
```

This was borrowed from WebRTC:
http://w3c.github.io/webrtc-pc/#data-transport-closing-procedure

There may be other ways to phrase it.

mfoltzgoogle:
I disagree.  I think the closed event should signal exactly that: the messaging channel is closed, i.e. no more messages can be accepted through the channel.  Pending requests to `send` may continue to be delivered but this would be best effort and at the discretion of the implementation.  For example, if the channel has abruptly closed (because the destination context went away, or the underlying transport was disconnected) no further delivery would be possible, and I think writing a spec to enumerate all the potential cases when messages could or could not be delivered would be complex.
 
I suggest the way forward is to land this PR and continue work in #240 on specifying a closing handshake, if it's an important use case to deliver messages to a closing/closed connection.

PR #248, opened 2016-01-28
Update security requirements for resolution around mixed content.

mfoltzgoogle:
Addressing Issue #80:

* Updates the security section with the resolution [1] around mixed content.
* Elaborates the circumstances in which an additional user agent can connect to a presentation.

[1] https://www.w3.org/2015/10/29-webscreens-minutes.html#action07

### DIFF
--- a/index.html
+++ b/index.html
@@ -878,10 +878,10 @@
             <li>Using the document's <a>settings object</a>, run the
             <a>prohibits mixed security contexts algorithm</a>.
             </li>
-            <li>If result of the algorithm is <code>"Prohibits Mixed Security
-            Contexts"</code> and <var>presentationUrl</var> is an <a>a priori
-            unauthenticated URL</a>, then return a a <a>Promise</a> rejected
-            with a <a>SecurityError</a>.
+            <li>If the result of the algorithm is <code>"Prohibits Mixed
+            Security Contexts"</code> and <var>presentationUrl</var> is an <a>
+              a priori unauthenticated URL</a>, then return a <a>Promise</a>
+              rejected with a <a>SecurityError</a>.
             </li>
             <li>If there is already an unsettled <a>Promise</a> from a previous
             call to <code>start</code> for the same <a>controlling browsing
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -862,6 +875,14 @@
...
             and abort these steps.
             </li>
+            <li>Using the document's <a>settings object</a>, run the
+            <a>prohibits mixed security contexts algorithm</a>.
+            </li>
+            <li>If result of the algorithm is <code>"Prohibits Mixed Security
+            Contexts"</code> and <var>presentationUrl</var> is an <a>a priori
+            unauthenticated URL</a>, then return a a <a>Promise</a> rejected
+            with a <a>SecurityError</a>.
+            </li>
```
tidoust: s/If result/If the result/
s/a a/a/

index.html
```diff
@@ -1001,6 +1022,14 @@
           <ol>
             <li>Let <var>P</var> be a new <a>Promise</a>.
             </li>
+            <li>Using the document's <a>settings object</a>, run the
+            <a>prohibits mixed security contexts algorithm</a>.
+            </li>
+            <li>If result of the algorithm is <code>"Prohibits Mixed Security
```
tidoust: s/If result/If the result/

index.html
```diff
@@ -2284,17 +2313,23 @@
...
       </p>
       <p>
-        This specification does not prohibit a user agent from publishing
-        information about its <a>set of presentations</a>. The group envisions
-        a user agent on another device (distinct from the controller or
-        presentation) becoming authorized to reconnect to the presentation,
-        either by user action or by discovering the presentation's URL and id.
+        This specification allows a user agent to publish information about its
+        <a>set of presentations</a>, and allows a browsing context on another
+        user agent connect to a running presentation via <code><a for=
```
tidoust: s/connect/to connect/

index.html
```diff
@@ -2284,17 +2313,23 @@
...
+        via a shared service.
+      </p>
+      <p>
+        However, the specification makes makes no guarantee as to the identity
+        of the connecting party. Once connected, the <a>receiving browsing
+        context</a> may wish to further verify its identity through
+        application-specific means. For example, the connecting <a>controlling
+        browsing context</a> may provide a token via <code><a for=
+        "PresentationConnection">send</a>()</code> that the <a>receiving
+        browsing context</a> can verify corresponds to a authorized entity.
```
tidoust: s/makes makes/makes/

As suggested by "application-specific means", it's not the "receiving browsing context" per se that may verify the identity of the connecting party, it's "the application running in the receiving browsing context".

Similarly, it's the "application running in the controlling browsing context" that may provide a token via send(), not the controlling browsing context itself.

I would suggest replacing occurrences of "browsing context" with "application" (or "party", perhaps?) in that sentence to emphasize that this check is up to the application.

mfoltzgoogle: Good point.  Reworded this section as suggested, and fixed the other grammatical errors noted.

Issue #79, opened 2015-04-20
Presentations from within nested browsing contexts


anssiko:
[Carved out into its own issue from https://github.com/w3c/presentation-api/issues/26#issuecomment-91708418]

@mfoltzgoogle says:

@avayvod  In Chrome have been discussing how browser-initiated
presentation interacts with multiple &lt;iframe&gt;s embedded in the
controlling page.  For now, we are only enabling browser-initiated
presentation from the default presentation URL set in the top level
document.  However, there is a use case for allowing it from an
embedded &lt;iframe&gt; for sites using a third party player to show a
single piece of video (or other) content.

We have discussed possible solutions, including:

1. Having the &lt;iframe&gt; pass a URL to the parent document to set as the
 default presentation URL.  The problem here is that the
browser-initiated presentation session would go to the parent
document, not the &lt;iframe&gt; that has the presentation control logic for
 the video player.

2. Picking an &lt;iframe&gt; DPU if none is set at in the parent.  This may
lead to surprising behavior for the user, since the user is expecting
to present content related to the parent document, which cannot
directly control the behavior of the &lt;iframe&gt;.

3. Allowing the user to choose from a list of default presentation
URLs in the document.  This complicates the UX for starting a
presentation and it will be difficult for the user to make a
meaningful choice (who provides the URL is an implementation detail of
 the webapp).

Instead we are considering adding a capability to the &lt;iframe&gt;'s
sandbox attribute, "allow-default-presentation", that would allow the
parent page to designate up to one iframe per document as the source
of the default presentation URL for browser initiated presentation.
This attribute would be respected only for &lt;iframes&gt; one level deep.

anssiko:
The proposal by @mfoltzgoogle is aligned with the model employed by the Fullscreen API [1]. It is a good idea to reuse a model that is in use unless there are issues with it. Personally, I'm not aware of any, but I haven't deep dived into the Fullscreen API.

To summarize the proposal (assuming I got it right ;-)):

* Define a new `allow-default-presentation` content attribute for the HTML `iframe` element. In practice, something like this:
```
<iframe src="https://example.org/video/1234" allow-default-presentation></iframe>
```
* This content attribute must be set in order for the browser to be able to initiate a presentation from https://example.org/video/1234
* This prevents e.g. content from third parties to go fullscreen without explicit permission. A concern shared with the Fullscreen API [2].

Questions:

* Should we require the same for any `startSession()` invocation from within `iframe`? How about `joinSession()`?

[...]

PR #249, opened 2016-01-29
Handling of nested browsing contexts.

mfoltzgoogle:
This PR updates the spec with the resolutions to Issue #79 (Presentations from within nested browsing contexts).  It:

- Patches the HTML5 spec to add an `allow-presentation` token to `<iframe sandbox=>`
- Provides UI guidance that the UA should tell the user what origin is requesting presentation.

It also removes links to a couple of resolved issues.

### DIFF
--- a/index.html
+++ b/index.html
@@ -335,6 +335,8 @@
         history</a></dfn>, <dfn><a href=
         "https://www.w3.org/TR/html5/browsers.html#sandboxing-flag-set">sandboxing
         flag set</a></dfn>, <dfn><a href=
+        "https://www.w3.org/TR/html5/browsers.html#active-sandboxing-flag-set">active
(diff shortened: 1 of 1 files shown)


anssiko:
Thanks for the pull! I committed some further clarifications to this branch:

* Defined "presentation is disabled" we can reference elsewhere in the spec
* Hooked that "presentation is disabled" into the entrypoint `navigator.presentation` (assuming the intended behaviour is to return null if presentation is disabled)
* Made the informative text in the Sandboxing section a note for clarity

@mfoltzgoogle Feel free to merge this PR, if you're happy with the clarifications.

mfoltzgoogle:
Thanks for the edits @anssiko.  Merging and closing issue.

PR #250, opened 2016-02-01
Fix #211: Define binaryType default value and get/set behaviour

### LINKED ISSUE #211: binaryType attribute's default value and get/set behaviour needs to be specified
I'm assuming the intended behaviour is similar to WebSocket, in which case can borrow its conventions. Specifically need to:

* Define the attribute's default value.
* Define how the attribute behaves on getting and setting (assume the default boilerplate will do: "On getting, it must return the last value it was set to. On setting, the user agent must set the IDL attribute to the new value." will do).

anssiko:
@mfoltzgoogle This is an attempt to Fix #211. Please review, and see if this matches your expectations. 

I defined "arraybuffer" as default since it seems to be a more future proof type that more and more APIs are able to consume. I also added a non-normative note borrowed from the WebSocket spec, mostly to clarify the behaviour to web developers.

### DIFF
--- a/index.html
+++ b/index.html
@@ -335,6 +335,8 @@
         history</a></dfn>, <dfn><a href=
         "https://www.w3.org/TR/html5/browsers.html#sandboxing-flag-set">sandboxing
         flag set</a></dfn>, <dfn><a href=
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
LGTM

PR #257, opened 2016-02-04
Complete Security and Privacy section

anssiko:
This PR attempts to fix the remaining open issues in the Security and Privacy section enumerated in issue https://github.com/w3c/presentation-api/issues/45#issuecomment-178288674. These issues were blocking the [wide review WD][1] publication.

We will continue to improve the Security and Privacy sections based on new information received from horizontal reviews, and can reopen issues #45 when new information resurfaces.

[1]: https://github.com/w3c/presentation-api/milestones/wide-review-WD

### DIFF
--- a/index.html
+++ b/index.html
@@ -2360,7 +2360,6 @@
       <h2>
         Security and privacy considerations
       </h2>
-      <div class="issue" data-number="45"></div>
       <h3>
         Personally identifiable information
       </h3>
--- a/index.html
+++ b/index.html
@@ -2373,6 +2373,15 @@
         However, this information is also dependent on the user's local network
         context, so the risk is minimized.
       </p>
+      <p>
+        The API enables <a href=
+        "#monitoring-the-list-of-available-presentation-displays">monitoring
+        the list of available presentation displays</a>. How the user agent
+        determines the compatibility and availability of a presentation display
+        with a given URL is an implementation detail. This feature can be used
+        to probe information about DIAL applications the user has installed on
+        the presentation display without user interaction.
+      </p>
       <h3>
         Cross-origin access
       </h3>
--- a/index.html
+++ b/index.html
@@ -2387,12 +2387,12 @@
       </p>
       <p>
         The API enables <a href=
-        "#monitoring-the-list-of-available-presentation-displays">monitoring the
-        list of available presentation displays</a>. How the user agent
+        "#monitoring-the-list-of-available-presentation-displays">monitoring
+        the list of available presentation displays</a>. How the user agent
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2497,10 +2505,12 @@
         The set of presentations known to the user agent should be cleared when
         the user requests to "clear browsing data."
       </p>
-      <div class="issue">
-        The spec should clarify what is to happen to the set of known
-        presentations in "incognito" (private browsing context) mode.
-      </div>
+      <p>
+        When in private browsing mode ("incognito"), the set of presentations
+        known to the user agent must be initially empty, and all state
```
mfoltzgoogle: I think this needs to be more specific - are we making a requirement of the controlling or receiving user agent?

What if I start a presentation from a non-incognito context - would I be allowed to connect to it from an incognito context (similar to connecting to it from another device)?

index.html
```diff
@@ -2373,6 +2372,15 @@
         However, this information is also dependent on the user's local network
         context, so the risk is minimized.
       </p>
+      <p>
+        The API enables <a href=
+        "#monitoring-the-list-of-available-presentation-displays">monitoring
+        the list of available presentation displays</a>. How the user agent
+        determines the compatibility and availability of a presentation display
+        with a given URL is an implementation detail. This feature can be used
+        to probe information about DIAL applications the user has installed on
```
mfoltzgoogle: Maybe, 

> If a <a>controlling user agent</a> can match a <a>presentation request URL</a> to a <a>DIAL</a> application and determine its availability, this feature can be used to probe information about which applications ...

Also, add a reference for DIAL?

http://www.dial-multiscreen.org/

anssiko: Good suggestion. Please feel free to update the PR.

index.html
anssiko: I recalled the receiving UA to be de facto in incognito, but notice that is not specified. Thus I was thinking controlling UA here initially. Now I realize this needs to be clarified. Feel free to amend the PR.

Re session started from non-incognito, then connected to from incognito. Wouldn't it be bad UX to deny this? How about asking the UA to explain to the user incognito in such a case may not work as expected?

mfoltzgoogle: Updated the PR to clarify that this applies to the controlling user agent.

I think stating that there are no presentation connections inherited from the existing (non-incognito) session is the right assertion.  If the user explicitly creates a connection via `start()`, or initiating a `defaultRequest`, that should be possible.  Such connections should go away when the session ends.

Updated the wording to try to capture this.

index.html

PR #264, opened 2016-02-11
Editorial changes to termination conditions

anssiko:
- Define "send a termination signal" and "send a termination confirmation signal"
- Add section "Reacting to a termination confirmation signal in a controlling browsing context"
- Misc editorial and tidy


mfoltzgoogle:
LGTM, thanks for the improvements @anssiko.

Issue #266, opened 2016-02-12
[meta] Presentation API Testing


anssiko:
When doing a larger change spanning across components it is often helpful to have a single issue for tracking the bigger change. This is such a "meta issue" for tracking the development of the test suite for the Presentation API.

**Tasks and related sub-issues:**

- [ ] Update initial `web-platform-tests` PR: w3c/web-platform-tests/pull/2432
- [ ] Split `interface Presentation` IDL into partials: #230
- [ ] @louaybassbouss to suggest more tasks for us :-)
- [ ] Publish test results: https://github.com/w3c/test-results/

All - @louaybassbouss is your test facilitator for the Presentation API spec and will be responsible for coordinating the test suite development. Please work with @louaybassbouss by contributing new test cases and reviewing existing tests. See the [Work Mode > Testing wiki][1] for more information on how to setup your own test environment.

Happy testing!

[1]: https://www.w3.org/wiki/Second_Screen/Work_Mode#Testing

louaybassbouss:
thx @anssiko. I will take the responsibility to keep the IDL-harness up-to-date with the spec. I will try to extract the IDL for each conformance class (controlling and receiving UA) programatically. 
I'd like also to ask Presentation API implementors to share with the group more information about their implementations e.g. which version of controlling UA, which presentation devices, limitations, etc. These information may help testers while writing or reviewing tests. 
For testing activities, I'd like to start with basic tests around the following  features:
* monitor screen availability.
* launch new presentation.
* reconnect to a running presentation.
* join a running presentation.
* presentation connection lifecycle (and transition between connection states as stated in the spec).
* messaging (send and receive messages in the formats stated in the spec).

If you like to open an new issue please use the label [testing][1]. 
I will let you know when the initial `web-platform-tests` PR is merged. 

[1]: https://github.com/w3c/presentation-api/labels/testing

anssiko:
Thanks @louaybassbouss. I updated the initial comment with the suggested tasks for basic tests, and will link to related issues when they appear.

anssiko:
The [initial PR has landed][1] to establish the test structure for the Presentation API, so you're now able to run the work-in-progress tests in-browser at:

http://w3c-test.org/presentation-api/

The Presentation API tests are located in the `w3c/web-platform-tests` repo at:

https://github.com/w3c/web-platform-tests/tree/master/presentation-api

(I added these links to the head of the spec and updated the first comment in this issue accordingly.)

[...]

PR #267, opened 2016-02-15
Split interface Presentation IDL into partials

### LINKED ISSUE #230: Presentation API IDL Tests for Presentation Interface
The Presentation API specifies interfaces for controlling and receiving user agents in the same spec. The [Presentation Interface][1] is one of the common interfaces used in both classes of user agents. In controlling UA only the `defaultRequest` attribute is relevant and for receiving UA only the `receiver` attribute. If one of these attributes is `undefined` (e.g. `defaultRequest` is not defined in receiving user agent or `receiver` is not defined in controlling user agent), the IDLharness will fail. I see 3 options to solve the problem:
1. Modify the Presentation Interface in the idlharness by remove the defaultRequest attribute in the idlharness of receiving UA and receiver attribute in the idlharness of controlling UA.
2. Implementors set `defaultRequest` to `null` in receiving UA and `receiver` to `null` in controlling UA (need to make this also clear in the spec). 
3. Define two different interfaces instead of one common Presentation Interface. 
 
I prefer option 2. What do you think?

[1]: http://w3c.github.io/presentation-api/#interface-presentation

anssiko:
* Split interface Presentation IDL into partials to make testing easier (fixes #230)
* Clarify the `defaultRequest` attribute returns null in a receiving browsing context

@mfoltzgoogle @louaybassbouss Please review.

### DIFF
--- a/index.html
+++ b/index.html
@@ -746,13 +746,10 @@
           partial interface Navigator {
             [SameObject] readonly attribute Presentation? presentation;
           };
-
+          
           interface Presentation {
-            attribute PresentationRequest? defaultRequest;
-            [SameObject] readonly attribute PresentationReceiver? receiver;
           };
-
-</pre>
+        </pre>
         <p>
           The <dfn for="Navigator"><code>presentation</code></dfn> attribute is
           used to retrieve an instance of the <a><code>Presentation</code></a>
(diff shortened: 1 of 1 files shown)


louaybassbouss:
@anssiko LGTM. In addition to your changes, I'd like to add for all IDLs the class `idl-controlling-ua` or `idl-receiving-ua` to make extraction (programatically) of IDLs for only controlling UA or receiving UA easier. If an IDL is required for both conformance classes (e.g. PresentationConnection), we can assign the two classes to the IDL at same time. What do you think?

anssiko:
@louaybassbouss Thanks for the review.

We should add such metadata if it makes test generation easier. I opened a [new issue][1] and updated the [meta issue][2] accordingly.

[1]: https://github.com/w3c/presentation-api/issues/268
[2]: https://github.com/w3c/presentation-api/issues/266

Issue #273, opened 2016-03-22
PresentationConnectionClosedEventInit.message needs to have a default value


smaug----:
...otherwise it isn't clear what the value of the following is:
(new PresentationConnectionClosedEvent("foo", { reason: "error" })).message

In general in dictionaries passed to an event constructor the properties need to have either default value or marked required.

mfoltzgoogle:
Closed by PR #277.

Issue #269, opened 2016-02-25
Only allow one PresentationAvailability per PresentationRequest


mfoltzgoogle:
The way the `getAvailability` algorithm is written [1], calling it multiple times on the same `PresentationRequest` will create multiple `PresentationAvailability` objects.  I think the intention is that there should be at most one created per `PresentationRequest`.

A small update to the algorithm should resolve this issue.

[1] https://www.w3.org/TR/presentation-api/#getting-the-presentation-displays-availability-information

mfoltzgoogle:
Closed by PR #277.

Issue #254, opened 2016-02-01
When is the first presentation connection created in the receiving browsing context?


tidoust:
The spec does not say anything about the creation of the first presentation connection in the set of presentations on the receiving side.

When the receiving browsing context is created, the `PresentationReceiver` singleton is created as well. Should the receiving user agent also create the first `PresentationConnection` instance at the same time? Also, what should its presentation connection state be?

mfoltzgoogle:
Closed by PR #277.

PR #277, opened 2016-04-14
Address issues #254, #269, #273

mfoltzgoogle:
This PR makes minor spec changes to address outstanding issues.

#254: When is the first presentation connection created in the receiving browsing context?
#269: Only allow one PresentationAvailability per PresentationRequest
#273: PresentationConnectionClosedEventInit.message needs to have a default value

### DIFF
--- a/index.html
+++ b/index.html
@@ -2493,14 +2493,23 @@
                 </li>
               </ol>
             </li>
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
@tidoust, I reworded the initialization of the presentation controllers monitor and the presentation controllers promise to make it clear that they are created once (on first connection and first getting).  I think this looks pretty good so going ahead and merging.

Issue #218, opened 2015-11-06
Possibility for a character to be interpreted differently depending on locale


tidoust:
Hi all,

[I'm raising this as an issue for tracking purpose, but I do not think that there is any issue in the end, so actually suggests to close it once people have reviewed it, unless someone points out that I missed something, of course!].

The group discussed potential internationalization issues during its F2F last week [1]. One particular issue that was raised was the possibility for a given character in a JS string to be interpreted differently by both browsing contexts, meaning it would be represented by different glyphs with different meaning in a Japanese, Chinese and/or Korean environment (e.g. a glyph meaning "Yen" in Japanese and a different currency in Chinese).

I said that I believed this was impossible in practice, and took an action "to investigate the possibility for a JS string to be rendered differently by different glyphs and locales".

I had a quick chat with @r12a (Richard Ishida), internationalization activity lead at W3C, and confirm that, unless I missed something in the scenario presented below, such a problem should never ever happen. The Presentation API operates on JavaScript types which are not affected by the character encoding used to retrieve the content.

The hypothetical scenario where a problem could have happened was something like:

1. the app running on the controlling browsing context is served encoded in Shift_JIS (usual encoding for Japanese characters);
2. the app running on the receiving browsing context is served encoded in Big5 (usual encoding for Chinese characters);
3. the app on the controlling browsing context extracts a string from a DOM element
4. the app on the controlling browsing context sends that string over to the receiving browsing context using the Presentation API's "send" method;
5. the app on the receiving browsing context sets the received value to a DOM element;
6. the characters rendered on the receiving browsing context for that DOM element mean something different.

In particular, regardless of the encoding used to serve a page/app, extracting a string from a DOM element returns a DOMString [2], which is an UTF-16 encoded serialization of the underlying sequence of Unicode characters (in an ideal world, this would return the sequence of Unicode character codes directly, but JavaScript strings are 16-bits only, so some characters are actually represented as two 16-bits surrogate pairs).

For instance, the Unicode character of a Japanese's "katakana letter small A" is 0x30A1, so if a DOM element contains such a letter, extracting it will yield a sequence with one integer 0x30A1, even if the document that was used to produce this element was encoded in Shift_JIS where this character is represented as a 0xA6 byte.

[...]

PR #280, opened 2016-04-19
Adds spec language around localization

mfoltzgoogle:
This PR implements the proposed actions from Issue #218.  Specifically it:

* Adds advice to Web authors for localized strings
* Adds example code for transmitting a localized string and inserting it into the presentation
* Adds a requirement to use language preferences to request presentation resources

### DIFF
--- a/index.html
+++ b/index.html
@@ -405,6 +405,11 @@
         in the File API specification [[!FILEAPI]].
       </p>
       <p>
+        The header <dfn><a href=
+        "https://tools.ietf.org/html/rfc7231#section-5.3.5">Accept-Language</a></dfn>
+        is defined in HTTP 1/.1 [[!rfc7231]].
+      </p>
+      <p>
         The term <dfn><a href=
         "http://w3c.github.io/webrtc-pc/#idl-def-RTCDataChannel">RTCDataChannel</a></dfn>
         is defined in the WebRTC API specification [[WEBRTC]].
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -648,6 +653,28 @@
...
 </pre>
       </section>
+      <section>
+        <h3>
+          Passing locale information with a message
+        </h3>
+        <pre class="example highlight">
+&lt;!-- controller.html --&gt;
+&lt;script&gt;
+  connection.send("{string: 'Hello, world!', lang: 'en-US'}");
```
tomoyukilabs: I would like to suggest including CJK example to more clearly indicate the case the locale could affect font rendering:

```
switch(navigator.language.toLowerCase()) {
  case: 'zh-cn':
    connection.send("{string: '你好，世界!', lang: 'zh-CN'}");
    break;
  case 'ja':
    connection.send("{string: 'こんにちは、世界!', lang: 'ja'}");
    break;
  case 'ko':
    connection.send("{string: '안녕하세요, 세계!', lang: 'ko'}");
    break;
  default:
    connection.send("{string: 'Hello, world!', lang: 'en-US'}");
    break;
}
```
(Note: This example includes all of CJK languages, however, one of them as well as English might be sufficient.)

tidoust: +1 to include a non-ASCII example. I don't think we need to bother with the `switch` clause to make our point, though:

```javascript
connection.send("{string: '你好，世界!', lang: 'zh-CN'}");
connection.send("{string: 'こんにちは、世界!', lang: 'ja'}");
connection.send("{string: '안녕하세요, 세계!', lang: 'ko'}");
connection.send("{string: 'Hello, world!', lang: 'en-US'}");
```

tomoyukilabs: @tidoust LGTM.

mfoltzgoogle:
Thanks @tomoyukilabs and @tidoust, I've added the extra examples you provided.  I'll go ahead and merge this PR.

PR #290, opened 2016-05-02
Pass the presentation identifier when starting a presentation.

mfoltzgoogle:
This change makes explicit that the presentation id should be passed from the controlling user agent when creating a presentation (receiving browsing context), and that the id of incoming presentation connection requests must match.


tidoust:
That looks good to me.

PR #292, opened 2016-05-05
Reject when sandbox blocks the feature instead of not exposing.

mounirlamouri:
The reasons are:
- navigator.presentation returning null only in this edge case will lead
  to bugs because developers will not expect this;
- we already have failure mode for all the methods;
- a malicious iframe could hold a Presentation instance if the embedder
  where to have the sandbox flags not set properly for a short period of
  time (eg. load race condition).

This change also does:
- add mixed security check for getAvailability() so we don't waste
  resources listen for availability when not needed;
- fix editorial issues whith mixed security check where the steps were
  not stopped.

### DIFF
--- a/index.html
+++ b/index.html
@@ -797,9 +797,8 @@
         <p>
           The <dfn for="Navigator"><code>presentation</code></dfn> attribute is
           used to retrieve an instance of the <a><code>Presentation</code></a>
-          interface. If <a>presentation is disabled</a>, the attribute MUST
-          return null. Otherwise, it MUST return the
-          <a><code>Presentation</code></a> instance.
+          interface. It MUST return the <a><code>Presentation</code></a>
+          instance.
         </p>
         <section>
           <h4>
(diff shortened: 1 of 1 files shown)


mounirlamouri:
Candidate follow-ups:
- Have a generic "Security check" so we don't need to repeat them three times.
- Say that `defaultRequest` should be ignored when the presentation is blocked from sandboxing.

mounirlamouri:
@mfoltzgoogle I'm happy to wait until the F2F if you think it's controversial. I think this is something we need to address for CR though (+ the defaultRequest issue).

index.html
```diff
@@ -1411,10 +1421,23 @@
...
-            disabled this feature), then:
+            <li>If one of the following conditions is true, run the next
+            substeps:
+              <ul>
+                <li>The user agent is unable to <a>monitor the list of available
+                presentation displays</a> for the entire duration of the
+                <a>controlling browsing context</a> (e.g., because the user has
+                disabled this feature).
+                </li>
+                <li>The result of running the <a>prohibits mixed security
```
mfoltzgoogle: Should this condition and the next reject P with SecurityError to match the behavior of start()?

index.html
```diff
@@ -1411,10 +1421,23 @@
...
             parallel</a>.
             </li>
-            <li>If the user agent is unable to <a>monitor the list of available
-            presentation displays</a> for the entire duration of the
-            <a>controlling browsing context</a> (e.g., because the user has
-            disabled this feature), then:
+            <li>If one of the following conditions is true, run the next
+            substeps:
+              <ul>
+                <li>The user agent is unable to <a>monitor the list of available
```
mfoltzgoogle: It seems like we should reject P with NotSupportedError here.  Instead of saying there are no displays, there's actually no way for the page to find out whether there are displays are not.

mfoltzgoogle: Perhaps the step @1451 can be merged here?

mfoltzgoogle:
Agree that defining behavior for Presentation is better than leaving it null.  Couple of questions about semantics of the APIs in these cases.

index.html
mounirlamouri: I was considering this option. Will apply the change.

index.html
mounirlamouri: Regarding resolving with `false` vs rejecting with `NotSupportedError`, I would suggest to deal with that in a follow-up because it's not related to this change.

However, IIRc, we did that so developers could take `NotSupportedError` as a clear signal that the discovery wasn't available but the feature would work otherwise.

mounirlamouri:
PTAL

mfoltzgoogle:
LGTM.  I think this improves the spec and is not potentially controversial, so merging, but leaving the issue open in case there are any comments between now and the F2F.

PR #298, opened 2016-05-19
Corrected Stop button ID in example code

chrisn:


avayvod:
lgtm as a typo fix
However, given that the actual `PresentationConnection` methods are called `terminate()` and `close()` maybe we should rename the buttons correspondingly so the example is a bit easier to follow?

Issue #99, opened 2015-06-01
Define behavior of Window and Fullscreen APIs in the presentation browsing context


mfoltzgoogle:
I think some normative statements about the behavior of these APIs should be considered. 

First, APIs that require input or handle focus should be discouraged from being supported:
* alert()
* confirm()
* blur()
* focus()
* print()
* prompt()

APIs that create new browsing contexts also be discouraged, as they are not friendly with 1-UA presentation.
* open()
* createPopup()

APIs that modify the position of the window are also not very meaningful in a reasonable implementation of the 1-UA mode:
* moveTo()
* moveBy()
* resizeTo()
* resizeBy()

Finally, we should support the  [Fullscreen API](https://fullscreen.spec.whatwg.org/), as web developers may want to use related CSS features as well as ensure fullscreen rendering of `<video>` content.  However, we may need to patch the Fullscreen spec for this scenario - need to dig further.

mounirlamouri:
I guess we could add the Presentation API in this list of APIs?

hsablonniere:
@mounirlamouri To prevent some kind of inception ? :smile:

hsablonniere:
@mfoltzgoogle I totally agree on the list of discouraged APIs but what would you enable from the CSS Fullscreen API?

Maybe we could have `:fullscreen` on the html element of the presentation browsing context but what about `::backdrop`? I'm not sure it makes sense to support it.

hsablonniere:
Isn't the presentation browsing context already displayed  in fullscreen?

mfoltzgoogle:
@mounirlamouri Well, the part of the API for receiving the PresentationSession is necessary for the page to communicate to the controller.

As for the controlling side, Jonas brought up the idea of a presentation "daisy chaining" onto additional screens to display alternate content.  See commit  https://github.com/w3c/presentation-api/commit/3917fd8da20d67f916bb4a711799d897dca85322 for an example use case.  But we don't yet have a concrete API for enabling daisy chaining or presentation to multiple displays with one user gesture.  So for now limiting the controlling side of the API in a presentation context seems reasonable.

Re: Fullscreen API

I am not sure we should guarantee the presentation is started in fullscreen.  The presentation display may be a PC running a windowing system, or I could forsee a side-by-side or picture-in-picture presentation modes for some displays.

Because the current fullscreen API spec does not allow content to go fullscreen without a user gesture [1], it will require some changes to support a presentation document fullscreening itself.

[...]

PR #304, opened 2016-05-26
Addresses APIs with user input on receiving user agent (Issue #99)

mfoltzgoogle:
This PR implements a proposed resolution for Issue #99; see https://github.com/w3c/presentation-api/issues/99#issuecomment-221434513

It adds a note to handle Window functions that require user input on the receiving user agent, especially when they block (i.e. alert())

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: node_js
+language: generic
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
Note that the commit list is long because of a merge from top of tree to an older branch.  (Not sure how to quash in GitHub)

PR #305, opened 2016-05-26
Addresses APIs with user input on receiving user agent (Issue #99)

mfoltzgoogle:
[Created a new branch to get a clean PR]

This PR implements a proposed resolution for Issue #99; see https://github.com/w3c/presentation-api/issues/99#issuecomment-221434513

It adds a note to handle Window functions that require user input on the receiving user agent, especially when they block (i.e. alert())

### DIFF
--- a/index.html
+++ b/index.html
@@ -2472,6 +2472,16 @@
             <a>databases</a> MUST be discarded and not used for any other
             <a>receiving browsing context</a>.
           </p>
+          <p class="note">
+            The <a>receiving user agent</a> may be a display-only device lacking
+            the ability to accept user input.  When this is the case, it should
+            handle functions that require user input carefully, especially when
+            the interaction is modal.  For example, functions such as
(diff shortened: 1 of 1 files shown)


mounirlamouri:
I would be a bit more generic and say that given that given the context of the receiving device, some APIs will not work per design (user input) or would become obsolete (window management, etc.). The UA needs to be aware of this. Furthermore, point that modal UI will need to be handled carefully.

mfoltzgoogle:
Thanks, please take a look at the revised wording @mounirlamouri.

anssiko:
The suggested wording looks good to me, thanks @mfoltzgoogle!

@mounirlamouri We'll interpret your silence as consent and land the revised PR.

mounirlamouri:
Sorry for not being responsive. I had a hard time to keep up with emails the last couple of weeks.

Issue #228, opened 2015-12-15
GitHub API failing, which leads to ReSpec failures


mfoltzgoogle:
Periodically, the calls to the GitHub API made inline in the spec document fail with error 403 (Forbidden) when executing ReSpec locally.  No other information is provided.    This causes ReSpec to silently stop processing the document.

I have commented out the githubAPI line in PR #227 until this issue can be sorted.

tidoust:
I believe that this is due to GitHub rate limiting unauthenticated requests (which is what ReSpec seems to be doing) to "60 requests per hour":
 https://developer.github.com/v3/#rate-limiting

I never run into problems when viewing specs online (my guess would be that GitHub does not apply the same limit when the request comes from code served from a github.io subdomain), but I confirm I've been running in similar issues from time to time when I work on a spec locally. I don't think there is a practical way to fix that in our case.

mfoltzgoogle:
Does every issue linked in the document cause a request to be sent?  I could see hitting this limit pretty quickly then.

tidoust:
No, I think there's only one (or two) calls to the API. The code that handles that part should be at:
https://github.com/w3c/respec/blob/develop/js/core/issues-notes.js#L130

Also, I note that conditional requests do not count towards the limit. However, I often force a refresh when I work locally to ensure I see the latest version of the spec.

PR #306, opened 2016-05-26
Issue #228: Remove GitHub issue code.

mfoltzgoogle:
Resolves Issue #228: GitHub API failing, which leads to ReSpec failures.  We have no more GitHub issues referenced in the spec, so we don't need this any more :-)

### DIFF
--- a/index.html
+++ b/index.html
@@ -108,9 +108,6 @@
             ]
           }
         },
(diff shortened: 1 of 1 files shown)


mounirlamouri:
lgtm

Issue #242, opened 2016-01-19
Describe the use cases for terminate vs close


birtles:
Not sure if this is the most appropriate place for feedback, but as a newcomer to the spec I'm having trouble understanding the use cases behind terminating vs closing. It seems that closing allows you to reconnect whereas terminate does not but I'm still puzzled as to what use cases they represent.

mfoltzgoogle:
We should ensure the use cases/reqs document [1] and examples [2] cover both of these use cases, and it's clear from reading the examples which use cases are demonstrated.  I'm pretty sure they do, but some clarifications are possible.

[1] https://github.com/w3c/presentation-api/blob/gh-pages/uc-req.md
[2] http://w3c.github.io/presentation-api/#examples

mfoltzgoogle:
@birtles Sorry it took a while to get back to this!  Please take a look at PR #307 and see if that clarifies things for you.  Thanks.

birtles:
Thanks, looks good.

PR #307, opened 2016-05-26
Add requirement and use cases for terminate presentation

mfoltzgoogle:
Addresses Issue #242: Describe the use cases for terminate vs close

This PR adds a requirement for terminating a presentation, and updates the use cases to show two scenarios where it would be used.

There are already use cases tagged with the requirement for signalling disconnection (i.e. close) so I did not add or update any use cases for it.


mounirlamouri:
lgtm

Issue #275, opened 2016-04-01
User Data Controls in Web Browsers guidelines


anssiko:
@travisleithead of TAG [reports][1]:

>At our most recent TAG f2f we discussed an alternate approach to our previously planned "privacy mode" spec. Given some pushback from implementers against standardizing the specifics of what constitutes privacy mode in browsers directly, @mnot has proposed a document that may be an incremental step toward that eventual goal. The document, aimed at spec authors, should provide a framework from which to consider the various user data controls that are made available in browsers. It is our hope that such a document could be a good reference for any spec concerned with how their feature is impacted by user privacy concerns. As the Second Screen group gave the TAG specific feedback about wanting to reference a privacy mode spec, we hope you'll take a look at this document and provide us feedback on whether it would be useful to you as a reference:
>
> [User Data Controls in Web Browsers][2]

Actions to the group:

- [x] Review the [User Data Controls in Web Browsers][2], provide feedback as comments in this issue
- [ ] Implement the guidelines in the Presentation API where applicable

Actions to @tidoust @anssiko:

- [ ] Summarize and submit the group's feedback to TAG

(For the context, the private mode was first discussed in the context of #45.)

[1]: https://github.com/w3ctag/spec-reviews/issues/61#issuecomment-204349002
[2]: https://gist.github.com/mnot/96440a5ca74fcf328d23
[3]: https://github.com/w3ctag/spec-reviews/issues/61to

tidoust:
I reviewed the [User Data Controls in Web Browsers](https://gist.github.com/mnot/96440a5ca74fcf328d23) document. I have three comments on the document itself:

1. It does not mention permissions exposed to the Permissions API in parts of the Web platform affected by Site Data Controls, which could persist from one browsing session to another. If it's on purpose, I don't understand why.
2. I failed to understand why AppCaches were listed in Site Data Controls and not in Local Data Controls together with HTTP caches initially, until I remembered that AppCache also comes with an Application cache API that exposes the cache status to the app. It might be worth clarifying that in the document. And if there's another reason, I missed it...
3. I'm not entirely clear whether we can reference the common user data controls identified in Appendix A, namely "Clearing Site Data" and "Privacy Mode", e.g. to say "Create a new top-level browsing context C set to display content on D in privacy mode" in an algorithm.

[...]

Issue #295, opened 2016-05-13
Check references category and stability


tidoust:
I'm creating this issue to track down references. It does not need to be addressed before publication as a Candidate Recommendation.

For the Presentation API to be published as a final W3C Recommendation, normative references will need to reference stable specs, the rule of thumb being that a Recommendation should normatively link to other Recommendations (or similar documents published by other organizations) and not to working drafts.

There may be exceptions to the rule, which should be documented in the Status of This Document section. For instance, since WebIDL is not yet a Recommendation, recently published specs have sentences such as "By publishing this Recommendation, W3C expects the functionality specified in this Recommendation will not be affected by changes to Web IDL as the WebIDL specification proceeds to Recommendation." as in the Web Storage spec:
http://www.w3.org/TR/2016/REC-webstorage-20160419/#status-of-this-document

Also, we need to ensure that references are correctly flagged as normative or informative, depending on how they get used in the spec.

Making a quick pass through the spec:
- [x] Check terms:
  - [x] "in parallel", defined in HTML5.1.
  - [x] WebIDL terms, although WebIDL Level 1 might move forward on the Rec track in the meantime
  - [x] `NotAllowedError`, defined in WebIDL (Second Level) but not in WebIDL Level 1
  - [x] `Blob`, defined in the File API spec
  - [x] Terms defined in the Mixed Content spec
  - [x] Terms defined in the Permissions spec
- [x] Convert to normative references (because of their use in the "create a receiving browsing context" algorithm):
  - [x] Cookies
  - [x] IndexedDB
  - [x] Permissions
  - [x] WebStorage
- [x] Consider updating the "create a receiving browsing context" to use prose such as "If the user agent supports IndexedDB, set the databases for C to an empty set of databases" to clarify that the action is conditional to the user agent supporting the feature in the first place.
- [x] Consider converting references to HTML 5.1 (Issue #396)

tidoust:
I believe the Presentation API now references stable published specifications whenever possible.

One remaining open question: can the Presentation API simply reference the latest version of WebIDL, even though it is not in a stable state, or should the Presentation API spec continue to reference both WebIDL Level 1 and WebIDL Level 2 to point out all WebIDL terms used in the spec are defined in Level 1 except `NotAllowedError` which is defined in WebIDL Level 2?

I'll try to come up with an answer (needed before requesting transition to Proposed Recommendation, not needed for publication as Candidate Recommendation).

mfoltzgoogle:
[...]

PR #308, opened 2016-05-27
Partially addresses Issue 275

mfoltzgoogle:
This PR implements parts of the resolution for Issue #275: User Data Controls in Web Browsers guidelines.    It also updates the list of site data to reflect implementation  status in the receiving user agent.

The parts remaining to be done include:

- Add a non-normative link to a reference list of site data when available.
- Add some text for Web Workers and Service Workers.  I would like to give this some more thought and possibly punt to separate issue.

### DIFF
--- a/index.html
+++ b/index.html
@@ -313,7 +313,9 @@
         Terminology
       </h2>
       <p>
-        The terms <dfn><a href="https://www.w3.org/TR/html5/browsers.html#application-cache">application cache</a></dfn>, <dfn><a href=
+        The terms <dfn><a href=
+        "https://www.w3.org/TR/html5/browsers.html#application-cache">application
+        cache</a></dfn>, <dfn><a href=
         "http://www.w3.org/TR/html5/browsers.html#browsing-context">browsing
         context</a></dfn>, <dfn><a href=
         "https://www.w3.org/TR/html5/browsers.html#nested-browsing-contexts">nested
@@ -428,9 +430,9 @@
         and is defined in HTTP/1.1 [[!rfc7231]].
       </p>
       <p>
-        <dfn><a href=
-        "https://tools.ietf.org/html/rfc7235#section-2">HTTP authentication</a></dfn>
-        is defined in HTTP/1.1: Authentication [[!rfc7235]].
+        <dfn><a href="https://tools.ietf.org/html/rfc7235#section-2">HTTP
+        authentication</a></dfn> is defined in HTTP/1.1: Authentication
+        [[!rfc7235]].
       </p>
       <p>
         The term <dfn><a href=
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -425,7 +427,12 @@
       <p>
         The header <dfn><a href=
         "https://tools.ietf.org/html/rfc7231#section-5.3.5">Accept-Language</a></dfn>
-        is defined in HTTP 1/.1 [[!rfc7231]].
+        and is defined in HTTP/1.1 [[!rfc7231]].
```
tidoust: Typo: drop "and"

index.html
```diff
@@ -2430,22 +2437,35 @@
...
-            <var>C</var> to <code>"denied"</code>.
+            <li>Set any state associated with <a>HTTP authentication</a> to the
+            empty state.
+            </li>
+            <li>Set the <a>application cache</a> to an empty <a>application
+            cache</a>.
             </li>
-            <li>Set the IndexedDB <a>databases</a> for <var>C</var> to an empty
-            set of <a>databases</a>.
+            <li>If the <a>receiving user agent</a> implements [[PERMISSIONS]],
```
tidoust: As raised in #295, this algorithm is normative, so references to external specs for actions should be normative as well (i.e. `[[!PERMISSIONS]]`). Same thing for references to IndexedDB and WebStorage below.

mfoltzgoogle:
Addressed comments by @tidoust.  Note that PERMISSIONS is a Working Draft, so it may need to be converted back to an informative reference (and the algorithm updated), depending on the outcome of #295.

index.html
```diff
@@ -2455,6 +2475,22 @@
...
+            <a><code>localStorage</code></a> MUST be discarded and not used for
+            any other <a>receiving browsing context</a>.
+          </p>
+          <p class="note">
+            The <a>receiving browsing context</a> is not intended to define a
+            private browsing session, but a well defined environment to allow
+            interoperable behavior for 1-UA and 2-UA presentations. For
+            example, navigation is not allowed in a <a>receiving browsing
+            context</a>.
+          </p>
```
tidoust: In the note, instead of "The receiving browsing context", I would rather use "This algorithm" (or "These steps").

Also, readers won't have the full context of WG discussions, and "private browsing session" comes a bit out of the blue here. I would simply strike that part and simply assert the intent.

Where do we say that navigation is not allowed in a receiving browsing context? The auxiliary navigation browsing context flag merely prevents the creation of additional browsing contexts, for instance. What the spec rather says is that the receiving user agent must terminate the presentation in case of navigation.

All in all, I would simplify the note to:
```
<p class="note">  
  This algorithm is intended to create a well defined environment to allow interoperable behavior for 1-UA and 2-UA presentations.
</p>  
```

tidoust:
Thanks @mfoltzgoogle. See additional comment on the note. The pull request looks good to me otherwise.

Not directly related to this pull request, but re-reading the algorithm, I think it does not really achieve what we'd like it to do. My understanding is that we'd like the user agent to create or use dedicated storages in the 1UA case. However:

1. The notion of *dedicated storages* (per browsing context) or even the notion of *storage* per se may not exist in the specs that are referenced. More precisely:
 1. There is only one cookie store per user agent in the cookies RFC, not one per browsing context. "Set the cookie store for C to an empty cookie store" does not mean much and implies emptying the cookie store of the controlling side as well in the 1UA case.
 2. IndexedDB databases are per origin, not per browsing context. "Set the IndexedDB databases for C to an empty set of databases" does not mean much and could be interpreted as requiring user agents to drop the contents of IndexedDB databases for all future connections, even those coming from regular (non receiving) browsing contexts. There is no notion of a storage that contains all IndexedDB databases in the IndexedDB spec.
 3. Unless I missed something, the HTML spec implicitly assumes that there is but one storage place for application caches. It does not explicitly define a term for that storage.
 4. Same thing for HTTP authentication state, where our goal is only to reset data for this browsing context, not for all browsing contexts.

2. Some properties are set by relevant algorithms during navigation, e.g. when the Document is created. Setting them to some empty storage beforehand won't do much, unless we also explicitly say that navigation should not affect these properties. Typically:
 1. sessionStorage and localStorage are set when the Document is created, which happens during navigation, c.f. two paragraphs below: http://www.w3.org/TR/webstorage/#dom-sessionstorage
 2. the application cache is selected during navigation or when the HTML document is parsed, c.f. one line below: http://www.w3.org/TR/html5/browsers.html#cache-host

Perhaps we can assume that the notion of storage is straightforward to grasp and tell the user agent to create new empty storages for the receiving browsing context in all cases, as in:

```
<ol>
  <li>Create a new <a>top-level browsing context</a> <var>C</var>,
  set to display content on <var>D</var>.</li>
  <li>Set the <a>session history</a> of <var>C</var> to be the empty
  list.</li>
  <li>Set the <a>sandboxed auxiliary navigation browsing context
  flag</a> on <var>C</var>.</li>
  <li>Create a new empty <a>cookie store</a> for <var>C</var>.</li>
  <li>Create a new empty store for <var>C</var> to hold
  <a>HTTP authentication</a> states.</li>
  <li>Create a new empty <a>application cache</a> storage for <var>C</var>.</li>
  <li>If the <a>receiving user agent</a> implements [[!PERMISSIONS]],
  set the <a>permission state</a> of all <a>Permissions</a> for <var>
    C</var> to <code>"denied"</code>.</li>
  <li>If the <a>receiving user agent</a> implements [[!INDEXEDDB]],
  create a new empty storage for IndexedDB <a>databases</a> for <var>C</var>.</li>
  <li>If the <a>receiving user agent</a> implements [[!WEBSTORAGE]],
  create a new empty storage for session storage areas and local storage
  areas for <var>C</var>.</li>
  <li><a>Navigate</a> <var>C</var> to <var>presentationUrl</var>.</li>
  <li>Start <a>monitoring incoming presentation connections</a> for
  <var>C</var> with <var>presentationId</var>.</li>
</ol>
```

I don't find that entirely satisfactory though. Any better idea?

anssiko:
@tidoust's proposed text looks pretty good to me (copy-paste-able proposals in HTML are nice :-)).

@mfoltzgoogle, given no concerns in 12 days after the @tidoust's proposal was aired, I suggest you update the PR with https://github.com/w3c/presentation-api/pull/308#issuecomment-222954662 and merge (after rebase & resolve merge conflicts).

anssiko:
To clarify, @tidoust noted Service Workers-provided dedicated storage (the `Cache` interface, any others?) is the only remaining unaddressed part in this PR.

To not block CR on the Service Workers considerations, we can note that in the SoTD as a WIP issue, unless we address this soon.

mfoltzgoogle:
I'm okay with the text suggested by @tidoust in https://github.com/w3c/presentation-api/pull/308#issuecomment-222954662 and https://github.com/w3c/presentation-api/pull/308#discussion_r65319128.  I think it's the best that can be done given the specs we are referring to here, and vendors will have to interpret the spirit of the spec going forward as new storage APIs get deployed.

mfoltzgoogle:
Adopted proposals from @tidoust with some minor edits to define local storage and session storage areas.  Merging.

Issue #293, opened 2016-05-05
defaultRequest text should explicitely mentions iframes that have presentation disabled from sandboxing


mounirlamouri:

anssiko:
ACTION @mounirlamouri to ship this before CR.

mfoltzgoogle:
Merged PR #317 with an updated version of the wording suggested in PR #310.

PR #310, opened 2016-05-30
Ignore defaultRequest if presentation disabled by sandbox flag.

mounirlamouri:


mounirlamouri:
@mfoltzgoogle PTAL.

mounirlamouri:
Fixes #293

mfoltzgoogle:
I might be more specific about what the UA should do if the sandboxing flag is set; as it reads, it seems like the controlling UA can do what it likes.

For example, something like:

> If the sandbox flag is set on the controlling context's document object, then the controlling UA should act as if the default request is not set for that browsing context.

anssiko:
@mounirlamouri to get this into the CR release, could you update the PR considering @mfoltzgoogle's feedback. Thanks!

Issue #311, opened 2016-05-31
Add presentation request URL fallback mechanism


anssiko:
[Discussed at the F2F, carving out into a separate issue.]

**PROPOSED RESOLUTION**: Add a fallback mechanism by adding a new constructor on `PresentationRequest` that takes an Array of DOMString, and by exposing the presented URL in an "url" attribute on `PresentationConnection`, pending feedback from implementers that this is doable.

For the full issue description and discussion, please refer the [F2F Day 2 minutes][1].

[1]: https://www.w3.org/2016/05/25-webscreens-minutes.html#item03

anssiko:
@mfoltzgoogle Did you receive feedback from your folks on the feasibility? This remains the only, and pretty essential, unresolved issue blocking our CR release.

mfoltzgoogle:
I did get feedback on this issue.  It will take 2-3 weeks of effort but should not significantly derail the timeline for us to get to CR.

As far as the PR to implement the resolution, @mounirlamouri offered to draft one, but he has been pulled into other urgent work.  After addressing other issues in this repo, I can take this on.

anssiko:
Fixed with #316.

PR #316, opened 2016-06-13
Updates PresentationRequest to take multiple URLs

mfoltzgoogle:
This PR addresses Issue #311 (Add presentation request URL fallback mechanism).  It modifies the PresentationRequest constructor to take an array of URLs each of which will be considered in order when starting or connecting to a presentation.

This necessitated several changes, including:
- Updating example code
- Updating interfaces, including adding a `url` property to PresentationConnection so the caller knows what is being presented
- Updating the set of availability objects to have a list of URLs per object
- Updating the list (set?) of available presentation displays to keep track of which displays are compatible with which URLs so the start algorithm can choose the appropriate URL for the chosen display
- Various other algorithm and wording updates

@mounirlamouri @tidoust @sicking @schien Please take a look.

### DIFF
--- a/index.html
+++ b/index.html
@@ -929,8 +929,8 @@
           Interface <a><code>PresentationRequest</code></a>
         </h3>
         <pre class="idl">
-          [Constructor(DOMString url)]
-          [Constructor(DOMString[] urls)]
+          [Constructor(DOMString url),
+           Constructor(DOMString[] urls)]
           interface PresentationRequest : EventTarget {
             Promise&lt;PresentationConnection&gt; start();
             Promise&lt;PresentationConnection&gt; reconnect(DOMString presentationId);
@@ -980,9 +980,8 @@
             </dd>
           </dl>
           <ol>
-            <li>If the constructor that take a single <var>url</var> was
-            invoked, let <var>urls</var> be a one item list with <var>url</var>
-            as its only member.
+            <li>If a single <var>url</var> was provided, let <var>urls</var> be
+            a one item array containing <var>url</var>.
             </li>
             <li>Let <var>presentationUrls</var> be an empty list of URLs.
             </li>
@@ -998,8 +997,8 @@
               </ol>
             </li>
             <li>Construct a new <code>PresentationRequest</code> object with
-            <var>presentationUrls</var> as the constructor argument and return
-            it.
+            <var>presentationUrls</var> as its <a>presentation request URLs</a>
+            and return it.
             </li>
           </ol>
         </section>
--- a/index.html
+++ b/index.html
@@ -940,7 +940,7 @@
         <p>
           When a <a><code>PresentationRequest</code></a> is constructed, the
           given <code>urls</code> MUST be used as the list of <dfn data-lt=
-          "presentation request URL|presentation request URLs">presentation
+          "presentation request URL">presentation
           request URLs</dfn> which are each a possible <a>presentation URL</a>
           for the <a><code>PresentationRequest</code></a> instance.
         </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -511,14 +511,15 @@
   // The Present button is visible if at least one presentation display is available
   var presentBtn = document.getElementById("presentBtn");
   // It is also possible to use relative presentation URL e.g. "presentation.html"
-  var presUrl = "http://example.com/presentation.html";
+  var presUrls = ["http://example.com/presentation.html",
+                  "http://another.com/alternate.html"];
```
tidoust: Examples should all use `example.com`, `example.net`, `example.org` or subdomains:
http://www.w3.org/2001/06/manual/#Examples

I suggest to switch to: `http://example.net/alternate.html`

index.html
```diff
@@ -917,7 +918,7 @@
           Interface <a><code>PresentationRequest</code></a>
         </h3>
         <pre class="idl">
-          [Constructor(DOMString url)]
+          [Constructor(DOMString[] urls)]
```
tidoust: Shouldn't we keep a `Constructor(DOMString url)` around as well that allows devs to pass only one presentation URL without having to wrap it in a one-item array? I realize that fallback URLs are going to be needed for some time, but I would keep the longer-term goal of "one presentation URL that works everywhere" in the API design (similar to the possibility to give only one "src" to a video tag)

index.html
```diff
@@ -1082,8 +1090,15 @@
...
-            <var>presentationUrl</var>.
+            <li>For each <var>presentationUrl</var> in
+            <var>presentationUrls</var> in order, run the following step:
+              <ol>
+                <li>If there is an entry <var>(presentationUrl, D)</var> in the
+                <a>list of available presentation displays</a>, then set the
+                <a>presentation URL</a> for <var>S</var> to
+                <var>presentationUrl</var> and stop iterating.
+                </li>
+              </ol>
```
tidoust: The above algorithm looks correct, but it's not obvious when you read it that it will **always** set the presentationUrl of `S`. Could the following make it clearer?

```
<li>Set the <a>presentation URL</a> for <var>S</var> to the first
<var>presentationUrl</var> in <var>presentationUrls</var> for which
there exists an entry <var>(presentationUrl, D)</var> in the <a>list of
available presentation displays</a>.
</li>
```

index.html
```diff
@@ -1500,13 +1521,27 @@
...
-            presentation display</a>, <code>true</code> otherwise.
+            object with its <code>value</code> property set as follows:
+              <ol>
+                <li>
+                  <code>false</code> if the <a>list of available presentation
+                  displays</a> is empty.
+                </li>
+                <li>
+                  <code>true</code> if there is at least one <a>compatible
+                  presentation display</a> for some member of of
```
tidoust: s/of of/of/

tidoust:
See a couple of comments inline. The only substantive comment is that I would have expected the `PresentationRequest` constructor that only takes one URL to remain in the spec.

Changes to algorithms look all good to me otherwise!

index.html

index.html
mfoltzgoogle: Agreed, better for backwards compatibility  as well.

index.html

index.html

index.html
```diff
@@ -928,7 +929,8 @@
           Interface <a><code>PresentationRequest</code></a>
         </h3>
         <pre class="idl">
-          [Constructor(DOMString url)]
+          [Constructor(DOMString url),
+           Constructor(DOMString[] urls)]
```
mounirlamouri: Do we want to allow (DOMString[] or DOMString) for simplicity and backward compatibility?

tidoust: That works for me. That seems to improve readability as well!

anssiko:
Thanks @mfoltzgoogle for crafting this substantial PR.

All - this PR has been open for 2 weeks, so my expectation is everyone has had time to review it. The comments received have been addressed by the editor sans @mounirlamouri's https://github.com/w3c/presentation-api/pull/316#discussion_r67614836 that suggested using a single `[Constructor]` with a union type `(DOMString[] or DOMString)` as its argument over the use of multiple `[Constructor]` extended attributes `[Constructor(DOMString url), Constructor(DOMString[] urls)]`.

@mounirlamouri could you clarify the compatibility issue?

Meanwhile, I'll merge merge this PR, with a condition that we may switch to the union type if the compatibility impact is deemed significant.

index.html
mfoltzgoogle: Sorry for the delayed reply, was on vacation last week.  The dual constructor form should be backwards compatible, although do I assume an implementation will be able to overload a single-argument constructor in this way?  For example code generated from the .idl would need to examine the argument type to dispatch to an appropriate function.

I'm not opposed to using a union type instead if there's a preference.

PR #326, opened 2016-07-20
Fix typo in code for Example 7.

lauweijie:
I believe `message` should be an EventHandler, so the message itself should be `message.data` as in the previous examples.


mfoltzgoogle:
LGTM.  Thanks @lauweijie

Issue #325, opened 2016-07-13
PresentationConnection.url should be optional?


mfoltzgoogle:
The algorithm steps for creating a PresentationConnection set the `url` attribute after construction.  It seems like we should make the attribute optional (like `id`) since it is not set in the constructor.

mfoltzgoogle:
Three algorithms that create PresentationConnection objects and set attributes after construction:

-  https://w3c.github.io/presentation-api/#starting-a-presentation
- https://w3c.github.io/presentation-api/#reconnecting-to-a-presentation
- https://w3c.github.io/presentation-api/#monitoring-incoming-presentation-connections

Alternatively, we could define a constructor that takes the url and id and modify the algorithms to use that; which seems better, since the API is not intended is to expose PresentationConnections without these properties set.

mfoltzgoogle:
Reviewing the spec, in all cases where the user agent instantiates a PresentationConnection,  it sets both the URL and the ID before returning it to the caller.  So it makes more sense to make both  `PresentationConnection.id` and `PresentationConnection.url` mandatory.

PR #332, opened 2016-08-18
Make PresentationConnection.id mandatory.

mfoltzgoogle:
Addresses Issue #325: PresentationConnection.url should be optional? 
See the comment there for the rationale for this change.

Since there is no way to get a usable PresentationConnection other than through the PresentationAPI, and the API sets the ID for all connections, I don't believe this change in semantics will affect any Web content.


tidoust:
I agree with the conclusion and the PR looks good to me.

Issue #315, opened 2016-06-12
list of displays: should it address display name, language, etc.?


aphillips:
The I18N working group is doing some housekeeping and found that we didn't forward the below [issue](https://github.com/w3c/i18n-activity/issues/113). Apologies for being late with this. Note that this may already be addressed by other threads about language.

http://w3c.github.io/presentation-api/

Section: The list of available presentation displays

This section contains the directive that:

>    The user agent MUST keep a list of available presentation displays. This current list of presentation displays may be used for starting new presentations, and is populated based on an implementation specific discovery mechanism. It is set to the most recent result of the algorithm to monitor the list of available presentation displays.

This list appears to be identified using URLs for each presentation display. The specification is silent about such issues as display name, sorting order, etc. These issues may affect actual implementation, even though this API doesn't really address the issue. Should it?

aphillips:
(could you please add the label "i18n" to this issue?)

tidoust:
I believe that the specification is silent on purpose here, to let implementers adjust their chrome and user experience as they see fit. For instance, a controlling user agent could choose to list displays that the user previously interacted with first, or sort displays it finds by proximity, or filter out displays that it deems not trustworthy, etc. In other words, the expectation is not that two controlling user agents will display the same list of available displays, and that they will appear in the same order.

The amount of information that the controlling user agent has on the available displays also depends on the discovery mechanism used under the hoods, which is out of scope of this specification. The user agent may be able to learn about all sorts of capabilities that the display supports, including locale support. Or it may not know much on top of the display name.

The spec could perhaps include some guidance about possible localization issues. I'm not clear how the display name could affect the actual implementation though. Do you have examples of issues that could arise, @aphillips ? For instance, is the problem that the user agent may not be able to render the name if the display name uses characters that are not supported by the user agent?

Side note: the list of displays is not identified using URLs. Each presentation display is associated with a list of URLs that it is compatible with, but it is not an identifier for the display. How the display gets identified is up to the implementation. It will probably be based on its IP address (possibly its MAC address on the LAN), or on some identifier exposed by the display itself.

aphillips:
@tidoust There are a number of issues that _could_ arise, but these would depend on implementation details. As long as the spec doesn't make normative statements about the presentation of the list, though, I'm not sure you need to enumerate what all of the potential problems are. Which is why the issue basically asks "should there be something here?". I'm not sure that there should be.

[...]

PR #333, opened 2016-08-23
Adds an implementation note for advertising and rendering a user friendly display name

mfoltzgoogle:
Addresses Issue #315: list of displays: should it address display name, language, etc.?

This adds an implementation note suggesting that receiving user agents advertise a user friendly display name with a locale and text direction, and controlling user agents render it using that information.


tidoust:
LGTM

tomoyukilabs:
LGTM, too.

Issue #303, opened 2016-05-26
Define that "TV" should appear as token for UA-string when presentation API is used to render on TV


sicking:
Mozilla's experience is that the few websites that today target TVs require that a "TV" token is included in the UA string in order for the website to render an appropriate TV UI.

I think this is something that's worth getting into a specification as quickly as we can, in order to prevent the mess that we currently have on mobile. The problem on mobile was that originally there was no token to indicate "mobile device", which lead to websites solving the problem with perpetually out-of-date databases, harder for new players to enter the space, etc.

At the very least, we should define that when implementing that 1-UA mode, the implementation should use a UA string apporpriate for the target device. So if using 1-UA mode to stream to a TV, the "hidden iframe" running on the controller device, should use a UA-string appropriate for TVs.

But I think it would also be good to include an informative note that when implementing the presentation-side of the presentation API on a TV, to include a "TV" token in the UA-string.

mfoltzgoogle:
This sounds like a workaround for a lack of CSS media queries for some document properties that might hold true when rendering on a presentation display.  What is a "TV" though?  Is it defined by screen size?  Pixel density?  Typical viewing distance? Input capability (or lack thereof)? Presence of HDMI inputs and a tuner?

Consider the following cases:
- Someone who uses a 81cm (32") or larger TV as a computer monitor on their desk
- A projector with smartboard or stylus input
- A laptop/PC connected to a TV by a wired HDMI connection as a second monitor

Would these cases count as "TV" for the purposes of User-Agent?  I'm not opposed to this outright, I just want to be as concrete as possible as to when it applies, and then understand if that would make sense to Web developers.

mfoltzgoogle:
This proposal would solve the problem that a document doesn't know for sure it's a presentation until the `connectionList` promise is resolved, and so would have to delay style decisions until that point.

If we simply say that the receiving user agent should include "TV" in the UA string for the receiving browsing context, that would seem to avoid tricky questions about "what is a TV."

tidoust:
Doesn't the presence of the `receiver` attribute on `navigator.presentation` convey the same information?

anawhj:
Most of smart TVs has the customized UA string for TV device, so I think the condition which determines whether it's TV or not, seems not to be defined.

For one thing, chrome browser on LG webOS TV has the userAgent string value as follows.
[...]

PR #350, opened 2016-09-27
Implements proposed resolution for Issue #303

mfoltzgoogle:
This addresses Issue #303: Define that "TV" should appear as token for UA-string when presentation API is used to render on TV.  It implements the proposed resolution: add a note about feature detection through the receiver attribute.


tidoust:
LGTM

Issue #340, opened 2016-09-16
Example 5.6 doesn't need onconnect handler


mfoltzgoogle:
The algorithm to monitor incoming presentation connections in a receiver will provide them to a page in a `connected` state.  Therefore the onconnect handler will never be called.

PR #352, opened 2016-09-27
Addresses Issue #340

mfoltzgoogle:
Addresses Issue #340: Example 5.6 doesn't need onconnect handler.  It removes the handler as connections are provided to the receiver in a connected state.


tidoust:
LGTM

Issue #341, opened 2016-09-16
navigator.presentation is marked optional in WebIDL


mfoltzgoogle:
In section 6.2, the presentation attribute is marked optional.  However the text says that it is mandatory for all conforming user agents to return an instance of Presentation.   I think this is just an IDL bug.

tidoust:
See [discussion at TPAC](http://www.w3.org/2016/09/22-webscreens-minutes#ins06)

**PROPOSED RESOLUTION:** For #341, make `navigator.presentation` attribute mandatory to align with the normative prose.

PR #353, opened 2016-09-27
Addresses Issue #341

mfoltzgoogle:
Addresses Issue #341: navigator.presentation is marked optional in WebIDL.  It implements the proposed resolution: make navigator.presentation attribute mandatory to align with the normative prose.


tidoust:
LGTM. Note to @louaybassbouss that IDL tests will need to be slightly updated as a consequence.

PR #358, opened 2016-09-30
Fix example 5.5 and move to ECMAScript2015

kenchris:
Fix errors in example 5.5 like missing reconnectBtn element and move to newer ECMAScript features where it makes sense.

### DIFF
--- a/index.html
+++ b/index.html
@@ -648,53 +648,72 @@
 &lt;!-- controller.html --&gt;
 &lt;button id="disconnectBtn" style="display: none;"&gt;Disconnect&lt;/button&gt;
 &lt;button id="stopBtn" style="display: none;"&gt;Stop&lt;/button&gt;
+&lt;button id="reconnectBtn" style="display: none;"&gt;Reconnect&lt;/button&gt;
 &lt;script&gt;
-  var connection;
+  let connection;
 
   // The Disconnect and Stop buttons are visible if there is a connected presentation
-  var disconnectBtn = document.getElementById("disconnectBtn");
-  var stopBtn = document.getElementById("stopBtn");
-  stopBtn.onclick = function () { connection &amp;&amp; connection.terminate(); };
+  const stopBtn = document.querySelector("#stopBtn");
+  const reconnectBtn = document.querySelector("#reconnectBtn");
+  const disconnectBtn = document.querySelector("#disconnectBtn");
 
-  var setConnection = function (theConnection) {
-    // Disconnect from existing presentation, if any
-    close();
+  stopBtn.onclick = _ => {
+    connection &amp;&amp; connection.terminate();
+  };
+
+  disconnectBtn.onclick = _ => {
+    connection &amp;&amp; connection.close();
+  };
+
+  function setConnection(newConnection) {
+    // Disconnect from existing presentation, if not attempting to reconnect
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -648,53 +648,75 @@
...
+&lt;button id="reconnectBtn" style="display: none;"&gt;Reconnect&lt;/button&gt;
 &lt;script&gt;
-  var connection;
+  let connection;
 
   // The Disconnect and Stop buttons are visible if there is a connected presentation
-  var disconnectBtn = document.getElementById("disconnectBtn");
-  var stopBtn = document.getElementById("stopBtn");
-  stopBtn.onclick = function () { connection &amp;&amp; connection.terminate(); };
+  let stopBtn = document.querySelector("#stopBtn");
```
mfoltzgoogle: Can you use const instead of let for the *Btn variables?

index.html
```diff
@@ -648,53 +648,75 @@
...
+
+  function setConnection(newConnection) {
     // Disconnect from existing presentation, if any
-    close();
+    if (connection) {
+      // Termination it not immediate, so run needed steps manually and remove callback
+      connection.onterminate();
+      connection.onterminate = undefined;
+
+      connection.terminate();
```
mfoltzgoogle: This could terminate the presentation when it's reconnected to, per example 5.3, if the old connection hasn't fired onclose yet.

If we want to guarantee that at most one connection to the presentation is open, then maybe this could be:

```
if (connection && connection != newConnection && connection.state != 'closed') {
  connection.onclosed = undefined;
  connection.close();
}
```

index.html
```diff
@@ -648,53 +648,75 @@
...
   // The Disconnect and Stop buttons are visible if there is a connected presentation
-  var disconnectBtn = document.getElementById("disconnectBtn");
-  var stopBtn = document.getElementById("stopBtn");
-  stopBtn.onclick = function () { connection &amp;&amp; connection.terminate(); };
+  let stopBtn = document.querySelector("#stopBtn");
+  let reconnectBtn = document.querySelector("#reconnectBtn");
+  let disconnectBtn = document.querySelector("#disconnectBtn");
+
+  stopBtn.onclick = _ => {
+    connection && connection.terminate();
```
mfoltzgoogle: I think you'll need `&amp;&amp;` here and below.

index.html
kenchris: sure

index.html
kenchris: Both seems to work, but let me make the change

index.html
kenchris: ok, incorporated

index.html
```diff
@@ -648,53 +648,72 @@
...
+    connection &amp;&amp; connection.terminate();
+  };
+
+  disconnectBtn.onclick = _ => {
+    connection &amp;&amp; connection.close();
+  };
+
+  function setConnection(newConnection) {
+    // Disconnect from existing presentation, if not attempting to reconnect
+    if (connection &amp;&amp; connection != newConnection && connection.state != 'closed') {
```
mfoltzgoogle: `&&` should be `&amp;&amp;`

mfoltzgoogle:
To move things forward I'll merge this and submit a follow-up pull request to make any necessary further changes.

Issue #359, opened 2016-10-14
"setting presentation.defaultRequest will have no effect" note is ambiguous


tidoust:
See related discussion between @schien and @tomoyukilabs on Web Platform Tests repository: https://github.com/w3c/web-platform-tests/issues/3844

In particular, @schien notes that:

> However, in the second note it says "setting presentation.defaultRequest" should take no effect. I take this sentence as "assigning value to defaultRequest will be no-op", which means presentation.defaultRequest will still be null after the assignment. That's my interpretation.

To which @tomoyukilabs replies:

> I'm afraid that the spec might look ambiguous. I regard it as "Although the defaultRequest property can be set to any instance of PresentationRequest or null, browsers which do not support defaultRequest will never refer to its value unless a script accesses it directly".

I would read the spec as @tomoyukilabs. Regardless of who is correct, it seems a good idea to clarify the expected behavior!

mfoltzgoogle:
The intent is that the attribute setting algorithm [1] completes, but there is no attribute-specific behavior for `defaultRequest` in the controlling user agent (step 9).

Since the spec does not override the attribute setting algorithm, the meaning seems clear enough to me, but the text could be explicit that the attribute is set.

[1] https://www.w3.org/TR/WebIDL/#dfn-attribute-setter

mfoltzgoogle:
There's another issue with starting a default presentation.   The spec says that the controlling user agent should run the steps to start a presentation (starting at step 2).  However, it's not clear that the user agent should construct a Promise as part of those steps (as there's nowhere to return it.)  Maybe we can adjust the language or construct a new algorithm that omits the Promise related steps.

louaybassbouss:
I am now thinking about the current behaviour of defaultRequest. In current spec the controlling page get notified (using `onconnectionavailable`) only if the user selects a display in the dialog. Do we need to inform the page that the selection dialog is open? in this case we need to change the spec as following:

```
navigator.presentation.defaultRequest = new PresentationRequest(presUrls);
navigator.presentation.defaultRequest.onstart = function(evt) {
    // here the page knows that a dialog is open
    evt.promise.then(setConnection).catch(...)
};
```

in this case we don't need to constrict a new algorithm for default request. What do you think?

mfoltzgoogle:
[...]

PR #369, opened 2016-11-18
Adds steps for initiating presentation from default presentation request.

mfoltzgoogle:
This PR addresses the issues noted in Issue #359 ("setting presentation.defaultRequest will have no effect" note is ambiguous).  Specifically:

1.  It clarifies that initiation of a presentation from the default request must come from the browser "chrome", not from any action of the page.
2. It lists specific steps for initiating a presentation from `start()` that check preconditions, allow the user to select a display, and return a Promise.
3. It lists specific steps for initiating a presentation from a default request that checks preconditions.
4. It clarifies that user agents that do not support initiation from a default request should ignore any value set for `navigator.presentation.defaultRequest.`

### DIFF
--- a/index.html
+++ b/index.html
@@ -1145,11 +1145,11 @@
                 </li>
               </ol>
             </li>
-            <li>If the user <em>denied permission</em> to use a display, reject
+            <li>If the user <em>denies permission</em> to use a display, reject
             <var>P</var> with an <a>NotAllowedError</a> exception, and abort
             all remaining steps.
             </li>
-            <li>Otherwise, the user <em>granted permission</em> to use a
+            <li>Otherwise, the user <em>grants permission</em> to use a
             display; let <var>D</var> be that display.
             </li>
             <li>Run the steps to <a>start a presentation connection</a> with
@@ -1293,7 +1293,7 @@
             </li>
             <li>Add <var>S</var> to the <a>set of controlled presentations</a>.
             </li>
-            <li>If <var>P</var> was provided, <a>resolve</a> <var>P</var> with
+            <li>If <var>P</var> is provided, <a>resolve</a> <var>P</var> with
             <var>S</var>.
             </li>
             <li>
--- a/index.html
+++ b/index.html
@@ -886,7 +886,7 @@
           In a <a>controlling browsing context</a>, the <dfn>default
           presentation request</dfn>, which is initially set to
           <code>null</code>, represents the request to use when the user wishes
-          to initiate a <a>presentation connection</a> from the browser.
+          to initiate a <a>presentation connection</a> from the browser chrome.
         </p>
       </section>
       <section>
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
Rawgit URL for review: https://rawgit.com/w3c/presentation-api/issue-359-default-request/index.html

PTAL @schien @tidoust

index.html
```diff
@@ -932,43 +932,22 @@
...
-          <div class="note">
-            Subsequent security steps of the <a>start a presentation</a>
-            algorithm apply. For instance, the request will fail if the
-            document object's <a>active sandboxing flag set</a> has the
-            <a>sandboxed presentation browsing context flag</a> set.
-          </div>
           <p>
-            Support for the initiation of a <a>presentation connection</a> from
-            the browser is OPTIONAL.
+            Support initating presentation using the <a>default presentation
```
tidoust: s/initating/initiating
Also, shouldn't that be "Support _for_ initiating"?

index.html
```diff
@@ -1170,9 +1151,123 @@
...
+            a presentation from a default presentation request</dfn>. If no
+            <a>default presentation request</a> is set on the document, these
+            steps MUST not be run.
+          </p>
+          <dl>
+            <dt>
+              Input
+            </dt>
+            <dd>
+              <var>W</var>, the document that the user intends to present
```
tidoust: I may be reading that incorrectly but, for me, the document that the user intends to present is the document that will be created on the receiving side when the presentation gets started. Use "the document on which the user expresses an intent to start presentation", instead?

index.html
```diff
@@ -1082,23 +1061,22 @@
...
           <h4>
-            Starting a presentation
+            Selecting a presentation display
           </h4>
           <p>
             When the <code><dfn for="PresentationRequest">start</dfn></code>
             method is called, the <a>user agent</a> MUST run the following
-            steps to <dfn>start a presentation</dfn>:
+            steps to <dfn>select a presentation display</dfn>, and then
+            <a>start a presentation connection</a> with the selected display.
```
tidoust: Although I don't see anyone could mis-interpret this and run the second algorithm twice, I would drop the end of the sentence.

In practice, the user agent will not run "select a presentation display", _and then_ "start a presentation connection". It will rather run "select a presentation display" which _includes_ a call to "start a presentation connection".

index.html
```diff
@@ -1208,6 +1306,8 @@
               The event must not bubble, must not be cancelable, and has no
               default action.
             </li>
+            <li>Let <var>U</var> be the the user agent connected to D.
```
tidoust: s/the the/the/

index.html

index.html

index.html

index.html
mfoltzgoogle: Yes, that's what I meant to say :) Updated text per your suggestion.

mfoltzgoogle:
Did you have any comments @schien?  Otherwise I plan to merge tomorrow Pacific time.

index.html
```diff
@@ -1170,9 +1150,124 @@
             all remaining steps.
             </li>
             <li>Otherwise, the user <em>granted permission</em> to use a
```
schien: Not sure why we use past tense here, "user grants permission" should be enough. Same thing in "user denied permission" few lines before.

index.html
```diff
@@ -1195,8 +1293,8 @@
             </li>
             <li>Add <var>S</var> to the <a>set of controlled presentations</a>.
             </li>
-            <li>
-              <a>Resolve</a> <var>P</var> with <var>S</var>.
+            <li>If <var>P</var> was provided, <a>resolve</a> <var>P</var> with
```
schien: Use "If P is provided"?

Issue #335, opened 2016-08-30
define the behavior for consecutive |getAvailability|


schien:
Three unclear behavior found while implementing the |getAvailability| algorithm:
1. In step 6, the same availability object is returned if an existed `(A, presentationUrls)` tuple is found. However, we can only return the same object if the requester is in the same DOM window.
2. in step 9, we should run the device monitoring algorithm in parallel so that we can resolve the promise as early as possible.
3. For consecutive |getAvailability| on the same `PresentationRequest` object, we should guarantee the resolve order in the algorithm. For example,

```
request.getAvailability().then(function() {console.log(1)});
request.getAvailability().then(function() {console.log(2)});
```

should always print "12" in console for all browsers.

tidoust:
See [discussion at TPAC](http://www.w3.org/2016/09/22-webscreens-minutes#ins11)

**PROPOSED RESOLUTION:** For #335, check algorithms around availability objects and make sure they all refer to "script object" or "internal object". For bullet 2, add "in parallel" in step 9. For bullet 3, make sure `getAvailability` returns the same `Promise` object.

mfoltzgoogle:
Section 6.4 will need a bit of cleanup to make it consistent with reasonable behavior.  Here is what I am thinking.
- Annotate `getAvailability()` with `SameObject`.
- Define a *presentation availability* internal object for use in the algorithms.  There is exactly one of these for a given list of URLs regardless of the script context.
- Define a *presentation availability promise* which is created at most once per *presentation availability*.
- Make sure that only one PresentationAvailability is created per script context per *presentation availability*.  (This is an optimization, may not be strictly required.)
- Update conditions when an entry can be removed from the *set of availability objects*.

Can start after the PR for Issue #342 lands (which tweaks some text in this same section).

PR #370, opened 2016-11-18
Clarifies the roles of objects for getting presentation display availability.

mfoltzgoogle:
This addresses Issue #335 by clarifying the roles of objects used by `getAvailability()`.

Specifically:
- Each `PresentationRequest` has at most one _presentation display availability_ and a _presentation availability promise_.
- The former is set to a `PresentationAvailability` object the first time it's needed.
- The latter is set to a new `Promise` the first time it's needed.

I did not use `[SameObject]` here as it did not seem appropriate per discussion in the other issue.

As far as I understand, with this clarification the example in https://github.com/w3c/presentation-api/issues/335#issue-173916158 should work by resolver order guarantees.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1624,10 +1624,10 @@
             <a for="PresentationAvailability">value</a> property of a
             <code>PresentationAvailability</code> object to offer presentation
             only when there are available displays. However, the <a>user
-            agent</a> may not support continuous availability monitoring; for
-            example, because of platform or power consumption restrictions. In
-            this case the <a>Promise</a> returned by <code><a for=
-            "PresentationRequest">getAvailability</a>()</code> is
+            agent</a> may not support continuous availability monitoring in the
+            background; for example, because of platform or power consumption
+            restrictions. In this case the <a>Promise</a> returned by
+            <code><a for="PresentationRequest">getAvailability</a>()</code> is
             <a data-lt="reject">rejected</a>, and the algorithm to <a>monitor
             the list of available presentation displays</a> will only run as
             part of the <a>select a presentation display</a> algorithm.
@@ -1785,11 +1785,12 @@
             Monitoring the list of available presentation displays
           </h4>
           <p>
-            If the <a>set of presentation availability objects</a> is non-empty,
-            or there is a pending request to <a for="PresentationRequest"
-            data-lt= "start">select a presentation display</a>, the <a>user
-            agent</a> MUST <dfn>monitor the list of available presentation
-            displays</dfn> by running the following steps.
+            If the <a>set of presentation availability objects</a> is
+            non-empty, or there is a pending request to <a for=
+            "PresentationRequest" data-lt="start">select a presentation
+            display</a>, the <a>user agent</a> MUST <dfn>monitor the list of
+            available presentation displays</dfn> by running the following
+            steps.
           </p>
           <ol link-for="PresentationAvailability">
             <li>Set the <a>list of available presentation displays</a> to the
@@ -1856,10 +1857,11 @@
             </li>
             <li>If the <a>set of presentation availability objects</a> is now
             empty and there is no pending request to <a for=
-            "PresentationRequest" data-lt="start">select a presentation display</a>,
-            cancel any pending task to <a>monitor the list of available
-            presentation displays</a> for power saving purposes, and set the
-            <a>list of available presentation displays</a> to the empty list.
+            "PresentationRequest" data-lt="start">select a presentation
+            display</a>, cancel any pending task to <a>monitor the list of
+            available presentation displays</a> for power saving purposes, and
+            set the <a>list of available presentation displays</a> to the empty
+            list.
             </li>
           </ol>
           <div class="note">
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
PTAL @tidoust @schien 

Rawgit URL: https://rawgit.com/w3c/presentation-api/issue-335-getavailability/index.html

I noticed that the text at L1518 that starts "However, the user agent may not support continuous availability monitoring" seems to conflict with step 5 of the algorithm that returns an availability object with `value` of false.  I think we should also update the text to match the algorithm.

index.html
```diff
@@ -1458,22 +1463,32 @@
             The set of availability objects
           </h4>
           <p>
+            Each <code>PresentationRequest</code> has a <a>presentation
+            availability promise</a> and a <a>presentation display
+            availability</a>, which are initially set to null. The
+            <dfn>presentation availability promise</dfn> for a
+            <code>PresentationRequest</code> is a <code>Promise</code> object
+            that <a data-lt="resolve">resolves</a> to a the <a>presentation
```
tidoust: s/a the/the/

index.html
```diff
@@ -1458,22 +1463,32 @@
             The set of availability objects
```
tidoust: Add "presentation" to the title as well to match the actual label of the set?

index.html
```diff
@@ -1458,22 +1463,32 @@
...
           </h4>
           <p>
+            Each <code>PresentationRequest</code> has a <a>presentation
+            availability promise</a> and a <a>presentation display
+            availability</a>, which are initially set to null. The
```
tidoust: s/null/<code>null</code>

index.html
```diff
@@ -1495,29 +1510,29 @@
...
+            While the <a>set of presentation availability objects</a> is not
+            empty, the <a>user agent</a> MAY <a>monitor the list of available
+            presentation displays</a> continuously, so that pages can use the
+            <a for="PresentationAvailability">value</a> property of a
+            <code>PresentationAvailability</code> object to offer presentation
+            only when there are available displays. However, the <a>user
+            agent</a> may not support continuous availability monitoring; for
+            example, because of platform or power consumption restrictions. In
+            this case the <a>Promise</a> returned by <code><a for=
             "PresentationRequest">getAvailability</a>()</code> MUST be
```
tidoust: That has not changed, but the "MUST" is already captured in the `getAvailability` algorithm. I would keep things informative here, e.g. with "in this case the Promise returned by getAvailability() is rejected".

index.html
```diff
@@ -1561,25 +1577,34 @@
...
+            availability promise</a> and abort these steps.
             </li>
-            <li>Return <var>P</var>, but continue running these steps <a>in
-            parallel</a>.
+            <li>Otherwise, set the <code>presentationRequest</code> object's
+            <a>presentation availability promise</a> to a newly created <code>
+              Promise</code> object and set <var>P</var> to this
+              <code>Promise</code>.
+            </li>
+            <li>Return the the <code>presentationRequest</code> object's
```
tidoust: s/the the/the/

index.html
```diff
@@ -1588,8 +1613,9 @@
               </ol>
             </li>
             <li>If the user agent is unable to continuously <a>monitor the list
-            of available presentation displays</a> but can find presentation
-            displays in order to start a connection, then:
+            of available presentation displays</a> at the current time but can
```
tidoust: I wonder whether we could further clarify the two "unable to monitor" steps:

1. I would switch them in the algorithm so that the step that rejects the promise appears first: it would make it clearer that a controlling browsing context that does not support background monitoring does not need to evaluate the second clause and will never resolve the promise with a newly created `PresentationAvailability` object in particular.
2. I'm not sure I understand "at the current time". Isn't "in the background" clearer, consistent with section 6.4, and what is meant here?

index.html
```diff
@@ -1632,10 +1662,10 @@
             </li>
             <li>Create a tuple <em>(<var>A</var>,
             <var>presentationUrls</var>)</em> and add it to the <a>set of
-            availability objects</a>.
+            presentation availability objects</a>.
             </li>
             <li>Run the algorithm to <a>monitor the list of available
-            presentation displays</a>.
+            presentation displays</a> in parallel.
```
tidoust: s/in parallel/<a>in parallel</a>

index.html

index.html

index.html

index.html

index.html
mfoltzgoogle: Both good suggestions and done.

mfoltzgoogle:
Applied changes suggested by @tidoust.  I also made a small change to the text I mentioned in https://github.com/w3c/presentation-api/pull/370#issuecomment-261663660, to clarify it applied to the case in the new step 5 (inability to monitor availability in the background).

mfoltzgoogle:
@tidoust, let me know if you have any further comments, I can address them in a followup PR.

Issue #360, opened 2016-10-17
Clarify event order when establishing a presentation connection


mfoltzgoogle:
In steps 19 - 23 of start a presentation [1] he intention of the spec (I believe) is that the following will happen in order:
1. P is resolved and the resolver function runs to completion.
2. The `connectionavailable` handler is invoked and runs to completion.
3. As part of the steps to establish a presentation connection [2], the `connect` handler is invoked or the `error` handler is invoked (and run to completion).

It occurs to me that since we haven't been explicit about the task sources for various asynchronous  steps the browser takes (like establishing a presentation connection) this order is implicit.  In our implementatation, a connection can complete in the background while step 20 is running for example.

Do we need to have stronger language in the spec around this @tidoust et al?

[1] https://w3c.github.io/presentation-api/#starting-a-presentation
[2] https://w3c.github.io/presentation-api/#dfn-establish-a-presentation-connection

tidoust:
Sorry I missed that issue somehow, @mfoltzgoogle.

I think it is a good idea to be explicit about the task source used. I would suggest the "networking task source" [1] (used by browser navigation WebRTC, WebSockets), unless there is a good reason to create a specific task source or to use more than one?

Wording could be: "The task source for the tasks mentioned in this specification is the networking task source".

This would guarantee that the task queued to fire the `connectionavailable` event (i) *runs* before the task queued to fire the `connect` event (ii).

However, I note that what guarantees that (i) gets *queued* before (ii) is not the task source: the steps are run "in parallel" so are not tied to the event loop, but "in parallel" steps are still expected to run one after another:
http://www.w3.org/TR/html51/infrastructure.html#in-parallel

Also, Promise resolution is not tied to any task queue, but rather run as a microtask, I think (before next task is processed).

[1] http://w3c.github.io/html/webappapis.html#networking-task-source

tidoust:
Upon further thought, I note that I'm not clear whether it is a good practice to reuse the "networking task source" or to create a new one. Do you know?

Creating a new task source would give implementers more flexibility on how they assign priorities to the tasks the spec creates compared to tasks created by other specs, and would still give us what we need, meaning proper relative ordering of the tasks.

I'm investigating internally. This does not change the fact that the spec needs to be explicit about the task source that gets used.

mfoltzgoogle:
[...]

PR #371, opened 2016-11-19
Clarify the task source for the specification.

mfoltzgoogle:
This addresses the comment in Issue #360 (Clarify event order when establishing a presentation connection).

### DIFF
--- a/index.html
+++ b/index.html
@@ -377,9 +377,10 @@
       <p>
         The terms <dfn><a href=
         "http://www.w3.org/TR/html51/infrastructure.html#in-parallel">in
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
PTAL @tidoust.

I didn't add any additional references to _in parallel_ in this PR.  If you see any steps where it would clarify the task order, let me know.

tidoust:
See question about whether we should rather create a dedicated task source in:
https://github.com/w3c/presentation-api/issues/360#issuecomment-262182053

This looks good to me otherwise, and you may want to merge this while we figure out whether it might be preferable to create a dedicated task source.

Issue #361, opened 2016-10-21
Use USVString for URLs


annevk:
The URL parser cannot handle lone surrogates.

PR #372, opened 2016-11-19
Replace DOMString with USVString for PresentationRequest URLs.

mfoltzgoogle:
Addresses Issue #361: Use USVString for URLs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1004,11 +1004,11 @@
           Interface <a><code>PresentationRequest</code></a>
         </h3>
         <pre class="idl">
-          [Constructor(DOMString url),
-           Constructor(sequence&lt;DOMString&gt; urls)]
+          [Constructor(USVString url),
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
PTAL @annevk who requested the change.

mfoltzgoogle:
Reverted the ID change.

(It seems unlikely that a unique ID generated by the browser will have a lone surrogate, but let's make that a separate issue.)

annevk:
Is the ID transmitted over the wire or serialized in some way? In that case USVString might make sense.

mfoltzgoogle:
Yes, it needs to be serializable as it can be sent to a separate receiving user agent over the network.  UTF-8 can encode lone surrogates, but we don't require that, so that sounds like a reason for using USVString to provide more flexibility.

annevk:
UTF-8 cannot encode lone surrogates. So yes, you should use USVString.

Issue #364, opened 2016-11-01
It is unclear what "If there is already an unsettled Promise from a previous call to start for the same controlling browsing context" means"


smaug----:
"If there is already an unsettled Promise from a previous call to start for the same controlling browsing context, return a Promise rejected with an OperationError exception and abort all remaining steps. "

Does that mean any PresentationRequest.start() call, or just if the same start() has been called before and promise is unsettled?

mfoltzgoogle:
The intention was to allow the user agent to use modal UI for display selection.  In that scenario, it would be complicated to support simultaneous requests, or confusing to service one request immediately after the other.

I'll craft a PR along the lines of,

> If there is already an unsettled Promise from a call to `start()` on any PresentationRequest in the same controlling browsing context, return a new `Promise` rejected with an `OperationError` exception and abort all remaining steps.

Alternatively, we could convert this step to an implementation note that browser is allowed to use modal UI for display selection and allowed to reject a simultaneous `start()` request according to the constraints of that UI.)

mfoltzgoogle:
@smaug--- if you have further feedback feel free to append it to this issue.  Otherwise I'll assume the PR addressed it.

PR #373, opened 2016-11-21
Clarify when a simultaneous call to start() can be rejected.

mfoltzgoogle:
Addresses Issue #364: It is unclear what "If there is already an unsettled Promise from a previous call to start for the same controlling browsing context" means"


mfoltzgoogle:
PTAL @smaug----

Issue #362, opened 2016-10-21
Consider checking for security context in the PresentationRequest constructor


mfoltzgoogle:
@mounirlamouri  brought up the following:

Currently, we check for mixed content in the steps to execute `start()`, `reconnect()` and `getAvailability()`.  It would simplify both the spec and the implementation to move this check to the PresentationRequest constructor.

I am not sure if it is okay to throw exceptions in conctructors of Web APIs.  If there are precedents, I'd like to know.

Also, we would need to be assured that the security context could not be changed between constructor and method invocation.

mounirlamouri:
There are precedent to throw in constructors. For example, the Fetch API does this for most of its defined constructor, for example [`Request`](https://fetch.spec.whatwg.org/#dom-request).

CC @annevk how does it sound?

annevk:
Seems fine. (I don't really understand how this specification works by the way. https://w3c.github.io/presentation-api/#getting-the-presentation-displays-availability-information says that there's some kind of input, whereas the method takes no arguments per IDL.)

Also, should this entire API be using `[SecureContext]`?

mfoltzgoogle:
Okay, let's move the security context check to the constructor.

The URL arguments come from the `PresentationRequest` that is the target of getAvailability().  Filed #363 to improve the description.

Can you provide a link to define `[SecureContext]`?

annevk:
Target? It's defined on that object, no?

SecureContext is defined in IDL.

schien:
@mfoltzgoogle Is this link you want? https://heycam.github.io/webidl/#SecureContext

mfoltzgoogle:
Note: I'm going to work on this after other PRs land that touch the same parts of the spec.

Re: `[SecureContext]` After security, privacy, TAG review #45 it was concluded to not restrict API to secure contexts.

annevk:
Can you point to the conclusion? I thought we wanted to move most new APIs to require them.

mfoltzgoogle:
Who is "we"?  One part of the relevant discussion is here.  This was also discussed with WebAppSec and PING, perhaps @tidoust can help dig up the relevant links.

https://www.w3.org/2015/10/29-webscreens-minutes.html#item06

annevk:
Mozilla at least. I also heard similar signals from @mikewest though Google overall might not be convinced yet.

But since you're going through a permission dialog it seems actually very important to require a secure context. Otherwise the user cannot be sure about who they are granting the permission to.

[...]

PR #377, opened 2016-11-28
Move mixed content check to PresentationRequest ctor.

mfoltzgoogle:
Addresses Issue #362.

### DIFF
--- a/index.html
+++ b/index.html
@@ -138,6 +138,7 @@
     table thead, table tbody { border-bottom: solid; }
     table td, table th { border-left: solid; border-right: solid; border-bottom: solid thin; vertical-align: top; padding: 0.2em; }
     dfn { font-weight: bolder; font-style: normal; }
+    var { hyphens: none; }
     .copyright { font-size: small; }
     .issue[id^='issue-'] > *:not([role='heading']) { display: none; }
     </style>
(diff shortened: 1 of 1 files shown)


mfoltzgoogle:
@mounirlamouri PTAL.

Note I could also move the sandboxing flag check to the ctor as well (removing it from the steps for start(), reconnect(), and getAvailability()). WDYT?

tidoust:
That looks good to me.

About the sandboxing flag, cannot sandboxing conditions change over time? For instance, the `sandbox` attribute of an `iframe` element may be removed at any time, which resets the sandboxing flag set:

> When an iframe element’s sandbox attribute is removed while it has a nested browsing context, the user agent must empty the iframe element’s nested browsing context’s iframe sandboxing flag set as the output.
> http://www.w3.org/TR/html51/semantics-embedded-content.html#element-attrdef-iframe-sandbox

Now, one could argue that once you have an `PresentationRequest` instance, you should be able to use it, even if sandboxing conditions change (and I don't see any good reason for an app to change the value of the `sandbox` attribute on the fly).

The check could also be done in the constructor **and** in start(), reconnect(), etc.

mounirlamouri:
lgtm, I think you should open an issue for the sandboxing flag. FWIW, a quick read of the spec let me think the sandboxing flags are set at navigation/creation but a quick read of the Blink code let me think it's updated dynamically.

Issue #363, opened 2016-10-25
Clarify arguments to getting presentation displays availability information


mfoltzgoogle:
Forked from #362.  It's not clear where the presentationUrls argument comes from when reading the steps in isolation.  Clarify the source of the input.

mfoltzgoogle:
Note: I'm going to work on this after other PRs land that touch the same parts of the spec.

PR #378, opened 2016-11-28
Clarifies inputs to algorithms.

mfoltzgoogle:
This addresses Issue #363: Clarify arguments to getting presentation displays availability information.

Specifically:

- Ensures that inputs to steps are written in terms of the objects that functions are called on and their arguments.
- Moves the precondition for the connection input to establishing a presentation connection steps into the steps themselves.
- Removes trailing punctuation throughout for consistency.


tidoust:
LGTM

Issue #379, opened 2016-11-29
Investigate if sandboxing flag check can move to the PresentationRequest ctor


mfoltzgoogle:
Follow-up from Issue #362.  It may be possible to check the sandboxing flag in the ctor for PresentationRequest, and this would simplify the algorithms that check it on function invocation.

However, if the attribute is dynamic or set on navigation, it may not be possible.

@mounirlamouri can you update this with what you discover in Blink?

mounirlamouri:
Looking at Blink code, it seems to be dynamic. My understanding of the spec is that it's not.

@annevk do I misinterpret the spec?

annevk:
I don't really understand the sandboxing setup. Too much forwarding of state.

mfoltzgoogle:
Well, it would be safest to check it in both the `PresentationRequest` ctor and its functions.

PR #385, opened 2016-12-01
Issue #379: Adds sandboxing flag check to PresentationRequest ctor.

mfoltzgoogle:
Addresses Issue #379: Investigate if sandboxing flag check can move to the PresentationRequest ctor.

For consistency, check the sandboxing flag in the `PresentationRequest` constructor.  Keep the checks in other functions the case the flag was unset at construction time and set later.

It may be that the sandboxing flag should be persistent, however, this does not seem to be implemented in Blink.  If that's true then we can remove the other checks later.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1043,6 +1043,10 @@
             </dd>
           </dl>
           <ol>
+            <li>If the document object's <a>active sandboxing flag set</a> has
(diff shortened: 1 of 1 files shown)


tidoust:
LGTM.
@louaybassbouss @tomoyukilabs note we'll need to update the sandboxing tests accordingly.

Issue #374, opened 2016-11-21
Determine if we need to fire a terminated event in the receiving browsing context(s)


mfoltzgoogle:
Currently, the steps for terminating a presentation don't fire a `terminated` event on connections held by the receiver.  If the process of closing the receiving browsing context could allow script to run, then we should consider firing this event (on the top-level or nested browsing contexts).  Doing so may be redundant with the 'beforeunload' event however.

Reference material:
- [Closing a browsing context](https://html.spec.whatwg.org/multipage/browsers.html#closing-browsing-contexts)
- [beforeunload event](https://html.spec.whatwg.org/multipage/browsers.html#beforeunloadevent)
- [unload event](https://html.spec.whatwg.org/multipage/browsers.html#unload-a-document)

mfoltzgoogle:
### Research notes

Using WhatWG [1] as a reference, there are three possible events that may fire during step 2 of presentation termination (unloading the presentation document).

- `onbeforeunload`
- `onpagehide`
- `onunload`

Comparing other device specs:

- As part of the "unload a document cleanup steps," the browser also "makes disappear" any open WebSocket objects [2], which closes the network transport but does not fire any events.
- A `MediaStreamTrack` may fire an `ended` event [3].
- A `RTCPeerConnection` does not clearly define what happens on unload. Presumably the user agent closes the peer connection, and those steps to close do not fire an event [4].
- A `BluetoothDevice` does not spell out an document unload procedure, but the disconnection algorithm does fire an event [5].

[1] https://html.spec.whatwg.org/multipage/browsers.html#prompt-to-unload-a-document
[2] https://html.spec.whatwg.org/multipage/comms.html#make-disappear
[3] https://www.w3.org/TR/mediacapture-streams/#life-cycle-and-media-flow
[4] https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-close
[5] https://webbluetoothcg.github.io/web-bluetooth/#clean-up-the-disconnected-device

mfoltzgoogle:
Noting that other specs are not very consistent in whether they event when closing resources on unload, and that firing a `terminate` event would be redundant with `beforeunload` for doing per-document cleanup, I'm inclined to *not* add a step to fire a `terminate` event on receiving presentation connections.

However, I do think we should set the presentation connection states to `terminated` before unloading the receiving browsing context so that state is available to `beforeunload` and other handlers.

Any other thoughts?

PR #386, opened 2016-12-01
Sets the state of receiving presentation connections to terminated before unload.

mfoltzgoogle:
Addresses Issue #374: Determine if we need to fire a terminated event in the receiving browsing context(s).

This changes the state of connections in the presentation to `terminated` before unloading the document in the receiving browsing context.  This is in lieu of firing an `terminate` event, which seems redundant with existing events like `beforeunload`.

Note that this change will prevent messages from being sent during `beforeunload`, which actually seems like it should simplify implementations:  they can disable and tear down the communication channels immediately upon receiving a signal to terminate.

PTAL

### DIFF
--- a/index.html
+++ b/index.html
@@ -2521,37 +2521,39 @@
             receiving browsing context</dfn>, it MUST run the following steps:
           </p>
           <ol>
-            <li>Let <var>P</var> be the presentation to be terminated.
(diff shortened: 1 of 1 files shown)


tidoust:
LGTM

Issue #387, opened 2016-12-01
Filling out the list of available presentation displays for start


tidoust:
Step 7. of the [select a presentation display algorithm](http://w3c.github.io/presentation-api/#dfn-select-a-presentation-display) tells the user agent to [monitor the list of available presentation displays](http://w3c.github.io/presentation-api/#dfn-monitor-the-list-of-available-presentation-displays). This is meant to fill out the [list of available presentation displays](http://w3c.github.io/presentation-api/#dfn-list-of-available-presentation-displays), which needs to be non empty for the display selection to do something.

The monitoring algorithm fills out that list based on the contents of the [set of presentation availability objects](http://w3c.github.io/presentation-api/#dfn-set-of-presentation-availability-objects). Problem is that set is only filled out by the [`getAvailability` algorithm](http://w3c.github.io/presentation-api/#getting-the-presentation-displays-availability-information).

In short, if an app calls `start` without first calling `getAvailability`, the monitoring algorithm will simply create an empty list of available presentation displays which would make `start` end up with a `NotFoundError` exception.

I guess the easiest way to fix that would be to have the selection algorithm set the presentation display availability right before current step 7, as done in `getAvailability`. Something like:

7. If the presentation display availability for presentationRequest is null, then:
  - Set the presentation display availability for presentationRequest to a newly created PresentationAvailability object, and let A be that object. 
  - Set the value property of A as follows: 
    1. false if the list of available presentation displays is empty. 
    2. true if there is at least one available presentation display for some member of presentationUrls. That is, there is an entry (presentationUrl, display) in the list of available presentation displays for some presentationUrl in presentationUrls. 
    3. false otherwise. 
  - Create a tuple (A, presentationUrls) and add it to the set of presentation availability objects.
8. If the user agent is not monitoring the list of available presentation displays, run the steps to monitor the list of available presentation displays in parallel.

Now, in theory, there should be no reason to introduce `PresentationAvailability` objects in `start`. Oh well...

tidoust:
Hmm, a better fix might be to complete the monitoring algorithm with a step such as:

If there is a pending request to select a presentation display, run the following step:
  1. For each display in newDisplays, if display is an available presentation display for one of the presentation urls then run the following steps:
     - let presentationUrl be that presentation url
     - Insert a tuple (presentationUrl, display) into the list of available presentation displays, if no identical tuple already exists.

mfoltzgoogle:
[...]

Issue #388, opened 2016-12-01
Check on Presentation display availability in getAvailability can never be true


tidoust:
One last one :) My apologies for not seeing that before as well!

[`getAvailability`](http://w3c.github.io/presentation-api/#getting-the-presentation-displays-availability-information) is the only algorithm that sets the [presentation display availability](http://w3c.github.io/presentation-api/#dfn-presentation-display-availability). That algorithm also creates (and never resets) the [presentation availability promise](http://w3c.github.io/presentation-api/#dfn-presentation-availability-promise).

The presentation display availability will always be `null` the first time `getAvailability` is called so condition in step 7 that checks on the presentation display availability will always evaluate to `false`. The presentation availability promise will always be set in subsequent calls to `getAvailability`, meaning that the algorithm will stop at step 2, return that promise, and never reach step 7.

In short, unless I missed something, I believe that step (and substeps) can be dropped.

This did not happen before because there was no such thing as a "presentation availability promise".

mfoltzgoogle:
Sorry, I don't follow.   The availability will be set to true if there is a display in the _list of available presentation displays_ that is compatible, found earlier by the monitoring steps.

tidoust:
Well, the step 7 I was looking at is now step 6: "If the presentation display availability for presentationRequest is not null, then: [...]".

The problem boils down to whether the user agent can reset the _presentation availability promise_ at some point in time, as I alluded to in a comment on PR #390: https://github.com/w3c/presentation-api/pull/390#issuecomment-264830700

If it cannot (which is how I read the current algorithm), then the promise will be set in any further call to `getAvailability`. That means any further call will stop at step 2, and so step 6. is useless.

Now, if the user agent can garbage collect the _presentation availability promise_ once it is settled and/or no longer referenced, then step 6. remains useful!

That step would also be useful if `start` can set the _presentation display availability_, as envisioned in PR #390.

mfoltzgoogle:
Okay, I am starting to understand what you are getting at.  Thanks for your patience.

The question is whether the availability promise/object is tied to the lifetime of the request that created them, which means they are created once and only once.  Or, if they can be GCed independently from the request and recreated later.

I believe the intention is that the developer can use the lifetime of the availability object to control when the browser is engaged in continuous monitoring of displays, and responsible usage would discard the object when the site isn't offering a UX to start presentation.

We didn't add a specific `discard()` or `dispose()` API for this; that doesn't seem to be a common Web idiom.

[...]

PR #390, opened 2016-12-03
Issue 387: Need to make sure we have a display list for start().

mfoltzgoogle:
Addresses issue #387: Filling out the list of available presentation displays for start.

If there is no existing _presentation display availability_ for the request, creates a temporary one and runs the steps to monitor.  The temporary one is removed at the end of `start()`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1129,19 +1129,24 @@
             <li>Return <var>P</var>, but continue running these steps <a>in
             parallel</a>.
             </li>
-            <li>Let <var>availabilityObj</var> be <code>null</code>.</li>
-            <li>If the <a>presentation display availability</a>
-                for <var>presentationRequest</var> is <code>null</code>, run the
-                following substeps <a>in parallel</a>.
+            <li>Let <var>availabilityObj</var> be <code>null</code>.
+            </li>
+            <li>If the <a>presentation display availability</a> for
+            <var>presentationRequest</var> is <code>null</code>, run the
+            following substeps <a>in parallel</a>.
               <ol>
-                <li>Set the <a>presentation display availability</a>
-                  for <var>presentationRequest</var> to a newly
-                  created <code>PresentationAvailability</code> object, and
-                  let <var>availabilityObj</var> be that object.</li>
-                <li>Create a tuple (<var>availabilityObj</var>, <var>presentationUrls</var>) and
-                  add it to the <a>set of presentation availability objects</a>.
+                <li>Set the <a>presentation display availability</a> for <var>
+                  presentationRequest</var> to a newly created
+                  <code>PresentationAvailability</code> object, and let
+                  <var>availabilityObj</var> be that object.
                 </li>
-                <li>Run the steps to continuously <a>monitor the list of available presentation displays</a>.</li>
+                <li>Create a tuple (<var>availabilityObj</var>,
+                <var>presentationUrls</var>) and add it to the <a>set of
+                presentation availability objects</a>.
+                </li>
+                <li>Run the steps to continuously <a>monitor the list of
+                available presentation displays</a>.
+                </li>
               </ol>
             </li>
             <li>Request user permission for the use of a <a>presentation
@@ -1168,7 +1173,8 @@
               </ol>
             </li>
             <li>If the user <em>denies permission</em> to use a display, reject
-            <var>P</var> with an <a>NotAllowedError</a> exception, and go to the step named <var>Cleanup</var>.
+            <var>P</var> with an <a>NotAllowedError</a> exception, and go to
+            the step named <var>Cleanup</var>.
             </li>
             <li>Otherwise, the user <em>grants permission</em> to use a
             display; let <var>D</var> be that display.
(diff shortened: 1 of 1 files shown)


tidoust:
I see why cleaning up the temporary _presentation availability display_ could be useful but, while stupid, cannot an app call `getAvailability()` immediately after calling `start()` and before the selection algorithm is over? `start()` should not reset _presentation display availability_ when it ends in that case because `getAvailability` needs it.

This triggers another issue, actually: when can the _presentation display availability_ be garbage collected? I read the current `getAvailability` algorithm as implying that the _presentation display availability_ (and the _presentation availability promise_) are set once and for all for the entire duration of the browsing context, and thus won't ever be garbage collected.

In turn, this would mean that the garbage collection steps envisioned in the [monitoring section](http://w3c.github.io/presentation-api/#monitoring-the-list-of-available-presentation-displays) will never run: "When a `PresentationAvailability` object is no longer alive (i.e., is eligible for garbage collection), the user agent SHOULD [...]"... because the availability objects would always remain "alive".

Am I reading this correctly? If so, the above GC steps need fixing. If not, I think the spec needs to be clearer that the _presentation display availability_ and the _presentation availability promise_ may be collected at any time (well, the promise must be collected before the availability object, since it contains a pointer to it).

mfoltzgoogle:
> I see why cleaning up the temporary presentation availability display could be useful but, while stupid, cannot an app call getAvailability() immediately after calling start() and before the selection algorithm is over? start() should not reset presentation display availability when it ends in that case because getAvailability needs it.

Is it possible for the `getAvailability()` steps to occur during the `start()` steps and before the _Cleanup_ step?

If so, we could have `getAvailability()` reject if there is an unsettled Promise from `start()`, or add a flag to indicate when the temporary availability object may cleaned up.

> This triggers another issue, actually: when can the presentation display availability be garbage collected? I read the current getAvailability algorithm as implying that the presentation display availability (and the presentation availability promise) are set once and for all for the entire duration of the browsing context, and thus won't ever be garbage collected.

Association of an availability object with a request shouldn't create a strong reference for the purposes of GC.  It's to prevent creation of duplicate objects.  How do specs label kinds of ownership of objects for GC purposes?

> Am I reading this correctly? If so, the above GC steps need fixing. If not, I think the spec needs to be clearer that the presentation display availability and the presentation availability promise may be collected at any time (well, the promise must be collected before the availability object, since it contains a pointer to it).

They should be collected like any other object, when they go out of scope for the associated script context's scope.  We should be more specific about the relationships among these objects so GC behavior is clear.

tidoust:
> Is it possible for the getAvailability() steps to occur during the start() steps and before the Cleanup step?

It seems possible at least given that steps run _in parallel_ each time.

> If so, we could have getAvailability() reject if there is an unsettled Promise from start()

It seems strange to tie the result of `getAvailability()` to `start()`.

> or add a flag to indicate when the temporary availability object may cleaned up.

Right, that would solve the problem.

> Association of an availability object with a request shouldn't create a strong reference for the purposes of GC. It's to prevent creation of duplicate objects. How do specs label kinds of ownership of objects for GC purposes?

I don't remember having seen strong/weak terminology used in other specs but maybe I missed something. It is not unusual to see garbage collection sections that describe when objects can be collected though. I do not know if there is an implicit default that the "internal slots" that a spec may use are weak references. It seems a good idea to be explicit about the intent in any case.

> They should be collected like any other object, when they go out of scope for the associated script context's scope. We should be more specific about the relationships among these objects so GC behavior is clear.

OK, then below is how I would fix this issue and clarify GC behavior. Happy to craft a PR if you think that's a good approach.

## In `start()`
Keep the algorithm as it exists today.

If you'd rather continue in the direction you set in this PR, I think you need to "synchronize" the `start` and `getAvailability` steps that set the _presentation display availability_ and fill out the _set of presentation availability objects_, otherwise you could in theory end up with duplicate entries in the set. It should be doable by shuffling things a bit to move these steps before the algorithms return the promise (and thus before algorithms start to run in parallel).

No clean up step should be needed.


## In `getAvailability()`
I think the intent with `getAvailability` is to return the same Promise as long as it is unsettled. To be confirmed with @schien (also raised in https://github.com/w3c/presentation-api/issues/388#issuecomment-264958260). If the goal is rather to return the same Promise as long as the code holds a reference to a Promise previously returned by that function, the suggestion below does not work.

I propose to get rid of the notion of _presentation availability promise_ entirely not to have to worry about specifying GC conditions for it:

1. replace current step 2 with: "If there is an unsettled Promise from a previous call to getAvailability on presentationRequest, return that Promise and abort these steps". This would make the method return the same Promise as long as it is unsettled.
2. adjust steps 3 and 4 accordingly


## In the monitoring algorithm
Start with a couple of steps that clone and extend the _set of presentation availability objects_ if needed:

1. Let _availabilitySet_ be a shallow clone of the _set of presentation availability objects_
2. If there is a pending request to "select a presentation display" for a presentation request and if the _presentation display availability_ of that presentation request is null, then create a tuple (A, presentationUrls) where presentationUrls are the presentation request URLs and add it to _availabilitySet_

Then use _availabilitySet_ in the rest of the algorithm.

An alternative approach would be to craft a separate set of steps for the pending request to "select a presentation display" as suggested in a previous comment (https://github.com/w3c/presentation-api/issues/387#issuecomment-264293136)

Also replace "When a PresentationAvailability object is no longer alive" with "When a _presentation display availability_ object is no longer alive"


## Next to the definition of presentation display availability
Add a sentence to say that user agents MAY collect the presentation display availability for a PresentationRequest at any time when the object goes out of scope for the associated script context's scope.

mfoltzgoogle:
Overall I like the approach you propose, here are a few things I noted.  Go ahead with a PR when you are ready.

> Keep the algorithm as it exists today.

Sounds good.

> I propose to get rid of the notion of presentation availability promise entirely not to have to worry about specifying GC conditions for it:

That would simplify things; but Web developers would need to make sure to expunge references to any Promises as well as the resulting availability object, in order for the availability object to be garbage collected.  This is a valid idiom for Promise-vending APIs, and I don't object to it.

> Start with a couple of steps that clone and extend the set of presentation availability objects if needed:

Since the monitoring steps run _in parallel_, if `getAvailability()` is called while it is running, we would just want to make sure that any new availability object created by those steps is inserted into the original _set of presentation availability objects_ and not the clone; and that the monitoring steps are run again to pick up the new object.

>  when the object goes out of scope for the associated script context's scope.

This is how Web Bluetooth puts it:

> When no ECMAScript code can observe an instance of BluetoothRemoteGATTServer server anymore, the UA SHOULD run server.disconnect(). 

Or we could just say "is eligible for garbage collection?"

It gets tricky because a script context can have multiple Realms / settings objects - so we would need to be specific about which one applies.   See Issue #391 I filed.

mfoltzgoogle:
This PR is obsoleted by PR #387.

PR #393, opened 2016-12-14
Rename PresentationConnectionClosedReason to PresentationConnectionCloseReason

foolip:
The name of an enum isn't observable, and this matches Blink's IDL. In
one place, "closeReason" is already used as a variable name.


mfoltzgoogle:
Thanks for noticing this and following up with a PR @foolip.

PR #395, opened 2016-12-19
Addresses Issue 387

mfoltzgoogle:
Addresses Issue #387: Well defined environment for nested contexts of the receiving browsing context?

This applies the restrictions placed on the receiving browsing context to child browsing contexts of that context.   It also clarifies that `navigator.presentation.receiver` should be `null` in those contexts.

Please check my terminology :)

### DIFF
--- a/index.html
+++ b/index.html
@@ -331,11 +331,9 @@
       <p>
         The terms <dfn><a href=
         "https://www.w3.org/TR/html5/browsers.html#application-cache">application
-        cache</a></dfn>, <dfn><a href=
+        cache</a></dfn>, <dfn data-lt="browsing context|browsing contexts"><a href=
         "http://www.w3.org/TR/html5/browsers.html#browsing-context">browsing
         context</a></dfn>, <dfn><a href=
-        "https://www.w3.org/TR/html5/browsers.html#nested-browsing-contexts">nested
-        browsing context</a></dfn>, <dfn><a href=
         "http://www.w3.org/TR/html5/webappapis.html#event-handlers">event
         handler</a></dfn>, <dfn><a href=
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2747,21 +2760,34 @@
             </li>
           </ol>
           <p>
+            All <a data-lt="nested browsing context">nested browsing
+            contexts</a> <a data-lt=
+            "creating a new browsing context">created</a> by the presented
+            document, i.e. that have the <a>receiving browsing context</a> as
+            their <a data-lt="ancestor browsing context">top-level
+            ancestor</a>, MUST also have restrictions 2-4 above. All of these
```
tidoust: Given the definition of top-level browsing context in HTML, I believe you can simply say "their <a>top-level browsing context</a>" here and drop the reference to "ancestor browsing context" (and drop it from the list of terms too).

index.html
```diff
@@ -380,7 +379,15 @@
...
         parallel</a></dfn> and <dfn><a href=
         "https://www.w3.org/TR/html51/webappapis.html#task-source">task
-        source</a></dfn> are defined in [[!HTML51]].
+        source</a></dfn>, <dfn><a href=
+        "https://www.w3.org/TR/html51/browsers.html#ancestor-browsing-context">ancestor
+        browsing context</a></dfn>, <dfn><a href=
+        "https://www.w3.org/TR/html5/browsers.html#nested-browsing-contexts">nested
+        browsing context</a></dfn>, and <dfn><a href=
+        "https://www.w3.org/TR/html51/browsers.html#list-of-the-descendant-browsing-contexts">
+        list of descendent browsing contexts</a></dfn>, and <dfn><a href=
```

index.html

Issue #380, opened 2016-11-29
Authenticity of screen selection permission is problematic in insecure contexts


mfoltzgoogle:
In Issue #362, @annevk raised the issue that the "authenticity" of the permission granting step for screen selection is problematic in insecure contexts.

@annevk can you provide any examples where this issue is addressed concretely in other specs/APIs?

annevk:
Well, for instance, Chrome started to require a secure context for geolocation.

annevk:
The overall problem is that users shouldn't have to pay attention to the lock icon when responding to dialogs. If they sometimes don't and sometimes do, you're just making things worse.

mfoltzgoogle:
Regarding geolocation, the motivation posted to blink-dev was that geolocation returned privacy sensitive information, and should not be exposed to a MITM, which I totally agree with.  We don't believe the Presentation API returns the same kind of privacy sensitive information.

The Secure Contexts TR [1] was also referenced to deprecate insecure use of "Powerful Features."  However "Powerful Features" are not defined by that spec.  @mikewest is this the document that defines what Chrome considers to be powerful features? [2] Is there anything in the W3C space that is the equivalent?

Regarding the dialog, that seems an issue with all Web platform APIs that require user interaction: alert(), file uploads, printing, running plugins, etc.  I'd like to understand better if the intention of some folks is to deprecate all of these on insecure contexts, and can follow up internally within Chrome.

[1] https://www.w3.org/TR/secure-contexts/
[2] https://www.chromium.org/Home/chromium-security/prefer-secure-origins-for-powerful-new-features

annevk:
The end goal is to get rid of insecure contexts, so yes.

mfoltzgoogle:
Chrome's guidelines on origin display are documented here [1].

In the origin display section of the spec [2], we can make it stronger to show that insecure origins are indeed insecure.

[1] https://www.chromium.org/Home/chromium-security/enamel#TOC-Presenting-Origins
[2] https://w3c.github.io/presentation-api/#user-interface-guidelines

annevk:
That doesn't really address my concern.

anssiko:
[Piggy-packing on this existing issue that provides the most recent context.]

[...]

PR #401, opened 2016-12-23
Adds a note about display of origins that are not potentially trustworthy.

mfoltzgoogle:
Addresses Issue #380: Authenticity of screen selection permission is problematic in insecure contexts.

This also alters the terminology to refer to potentially trustworthy origins, to align with the current drafts of Mixed Content [1] and Secure Contexts [2].

[1] https://www.w3.org/TR/mixed-content/
[2] https://www.w3.org/TR/secure-contexts/

### DIFF
--- a/index.html
+++ b/index.html
@@ -518,13 +518,17 @@
       </p>
       <p>
         The terms <dfn><a href=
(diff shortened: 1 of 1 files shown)


tidoust:
That works for me.

PR #410, opened 2017-02-15
Update Service Worker references

tomoyukilabs:
This PR migrates Service Workers references from old Editor's Draft to [Service Workers 1 Working Draft](https://www.w3.org/TR/service-workers-1/).


mfoltzgoogle:
Thanks @tomoyukilabs !

Issue #408, opened 2017-01-25
Reference to WebIDL


tidoust:
This issue is a follow-up to issues #295 (Check references category and stability) and #407 (drop [SameObject] for promise attribute).

Per process, normative references in the Presentation API should switch to stable specifications en route to Proposed Recommendation. At the time of writing, WebIDL Level 1 is a Recommendation, WebIDL Level 2 is an Editor's Draft. Reference should in theory be to Level 1 (except for NotAllowedError, defined in Level 2), as in the current draft of the Presentation API.

However, there are a number of known issues with Level 1. Can the Presentation API only reference Level 2 or WebIDL without level?

mfoltzgoogle:
Not sure who can answer this question?

anssiko:
@mfoltzgoogle, I think what @tidoust wanted to ask from the technical perspective is:

Does the WebIDL Editor's Draft at https://heycam.github.io/webidl/ (aka "WebIDL without level") have all the definitions the Presentation API needs?

Based on my assessment that is the case, so here's a proposed fix, PR #411.

I believe @tidoust has an open action via https://github.com/w3c/presentation-api/issues/295#issuecomment-272212292 to ensure referencing WebIDL Editor's Draft like this will not block the Proposed Recommendation publication (when we get there). Regardless, referencing "WebIDL Editor's Draft" will not block the update to the Candidate Recommendation tracked in #406.

(Confusion probably arises from the fact that the [WEBIDL-2] reference in the biblio database points to the "WebIDL Editor's Draft" at https://heycam.github.io/webidl/. However, there is no formal "WebIDL Level 2" spec.)

PR #411, opened 2017-02-15
Fix WebIDL references

anssiko:
Proposed fix to #408.

@mfoltzgoogle @tidoust PTAL.

### DIFF
--- a/index.html
+++ b/index.html
@@ -490,7 +490,7 @@
       </p>
       <p>
         This document provides interface definitions using the Web IDL standard
-        [[!WEBIDL-1]].
+        [[!WEBIDL-2]].
       </p>
       <p>
         The terms <dfn><a href=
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -500,8 +500,8 @@
         <dfn><a href=
         "https://heycam.github.io/webidl/#idl-ArrayBuffer">ArrayBuffer</a></dfn>,
         <dfn><a href=
-        "https://heycam.github.io/webidl/#idl-ArrayBufferView">ArrayBufferView</a></dfn>,
-        and the following exception names are defined in [[!WEBIDL-1]]:
+        "https://heycam.github.io/webidl/#ArrayBufferView">ArrayBufferView</a></dfn>,
+        and the following exception names are defined in [[!WEBIDL-2]]:
```
mfoltzgoogle: I don't think this phrase is required anymore since the entire spec is using WEBIDL-2?

anssiko: Can you clarify what you think is not required?

We should enumerate all `<dfn>`s the spec depends on (for completeness and for ReSpec cross-referencing to work), and by convention indicate the references from where they originate from. There's no one true way, so editor has some freedom whether to use a single bulleted list or break definitions per reference, like is done here.

mfoltzgoogle: LGTM

Okay, on reading the full spec this list is separate from the "interface definitions" statement above, so deserves its own cross reference.

Issue #414, opened 2017-02-25
Receiving browsing context needs additional flags set


mfoltzgoogle:
The receiving browsing context sets the _sandboxed auxiliary navigation browsing context flag_ to prevent it from using e.g. window.open() to spawn new browsing contexts.

However we should also prevent top-level navigation (to protect integrity of the screen selection prompt) by setting the _sandboxed top-level navigation browsing context flag_.  This also blocks `window.close()`, thus developers will have to call .terminate() for a presentation to close itself.

We can also set  the _sandboxed modals flag_ to prevent modal dialogs, which is suggested in a non-normative note, but not enforced.  I don't feel as strongly about this however.

anssiko:
Setting _sandboxed top-level navigation browsing context flag_ is indeed required to ensure integrity. The presentation is effectively a single-page application that must not be able to navigate. This should have been caught up in the security review, but better late than never. Good catch!

Looking at the _sandboxed modals flag_, it prevents content from using the following:

* window.alert()
* window.confirm()
* window.print()
* window.prompt()

Also the following, but IIUC, not relevant here:

* window.showModalDialog() - This is already disallowed by the _sandboxed auxiliary navigation browsing context flag_ we set, and also not anymore in the _sandboxed modals flag_ list in HTML LS, so probably a spec bug in 5.1).

* the beforeunload event - In _terminate a presentation in a receiving browsing context_ we _unload a document_, thus this event is never fired when a presentation is terminated (the beforeunload event is only fired when a user agent is to _prompt to unload a document_).

@mfoltzgoogle @schien How do your implementations handle alert(), confirm, print(), or prompt()? How about we specify the implemented behaviour if it is consistent among known implementations.

mfoltzgoogle:
In Chrome, all Javascript dialogs are suppressed from offscreen tab (1-UA) presentations.

schien:
In Firefox, open/alert/confirm/print/prompt will fail in receiver page, however, window.close will close the connection but not terminate the receiver page.

anssiko:
Thanks @schien and @mfoltzgoogle. Do your implementations also prevent top-level navigation in the receiving browsing context already?

[...]

PR #415, opened 2017-03-01
Sets top-level navigation and modal dialog flags on presentations.

mfoltzgoogle:
Addresses Issue #414: Receiving browsing context needs additional flags set.

It sets the _sandboxed top-level navigation browsing context flag_ and _sandboxed modals flag_ on the receiving browsing context, to protect the integrity of the origin being presented, and to reflect implementation behavior (pending Mozilla feedback).

It also drops a reference to calling `window.close()`, as this is not allowed with the former flag set.  Instead, the presentation must call `connection.terminate()` on any `PresentationConnection`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -440,6 +440,11 @@
         </li>
         <li>
           <dfn><a href=
+          "https://www.w3.org/TR/html51/browsers.html#sandboxed-modals-flag">
+          sandboxed modals flag</a></dfn>
+        </li>
+        <li>
+          <dfn><a href=
           "https://www.w3.org/TR/html51/browsers.html#sandboxing-flag-set">sandboxing
           flag set</a></dfn>
         </li>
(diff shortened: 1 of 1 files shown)


anssiko:
LGTM. Thanks!

index.html
```diff
@@ -2610,9 +2615,8 @@
           <ol>
             <li>The <a>receiving user agent</a> is to <a>unload a document</a>
             corresponding to the <a>receiving browsing context</a>, e.g. in
-            response to a call to <code>window.close()</code> in the
-            <a>top-level browsing context</a> or to a request to
-            <a>navigate</a> that context to a new resource.
+            response a request to <a>navigate</a> that context to a new
```
tidoust: s/in response a request/in response to a request

PR #417, opened 2017-03-23
Adds notes related to sandboxing flags on presentations 

mfoltzgoogle:
Followup to #415.  Adds notes to Create a Receiving Browsing Context and User Interface Guidelines around sandboxing flags.


tidoust:
Thanks, LGTM!

Issue #421, opened 2017-04-12
Relax `terminate` algorithm to deal with other states?


tidoust:
The `terminate` method only has an effect on a presentation connection whose state is `connected`.

I think a common scenario for calling `terminate` is when the controlling browsing context is about to disappear or when the user clicks on a button, to clean things up, regardless of the current application state.

This is what we try to do in the test suite for instance, so that subsequent tests start from a clean "no on-going presentation" state. However, since `terminate` only works on `connected` presentations, terminating a presentation no matter what turns out to be more complex that simply calling `terminate`, leading to code such as:

```js
// Try to terminate the presentation
connection.terminate();

// If state was "connecting", previous call did nothing,
// wait for the "connect" event
connection.onconnect = () => { connection.terminate(); };

// If state was "closed", we need to reconnect first. This will return a
// presentation in "connecting" state. Previous event handler will eventually
// take care of it if it's the same "connection", but it may not be.
if (connection.state === 'closed') {
  request.reconnect(connection.id).then(c => {
    c.onconnect = () => { c.terminate(); };
  });
}
```

I'm wondering whether the user agent could take care of some of that complexity, ideally so that developers can end up with a simple `connection.terminate()` call. This would require relaxing the `terminate` algorithm so that:

1. It also applies to connections in the `connecting` state. The user agent would wait for the underlying connection to become `connected`.
2. It also applies to connections in the `closed` state. The user agent would attempt to reconnect the underlying connection before it terminates it.

The latter change may be more problematic, and it seems reasonable to expect applications to deal with connections in the `closed` state themselves. The former change seems easier and useful. If both changes are in place, developers can just issue a call to `connection.terminate()` and be done with it. If only the former change is in place, developers would have to write something like:

```js
if (connection.state === 'closed') {
  request.reconnect(connection.id).then(c => { c.terminate(); });
}
else {
  connection.terminate();
}
```

... which seems acceptable.

mfoltzgoogle:
If I'm not opposed to this change, but would need to investigate the implications for our implementation to determine the scope and complexity.  It does seem to improve ergonomics for the cleanup case, but isn't the expected use case to simply allow the connection to be closed by the controlling UA on page unload?  The presentation can close itself when it detects the connection has goneAway.

I went ahead and thought through the implications for the API and what we signal to developers.

[...]

PR #427, opened 2017-05-02
Allow termination via connections in a connecting state.

mfoltzgoogle:
Addresses Issue #421: Relax `terminate` algorithm to deal with other states?

This allows termination of a presentation by a connection in the `connecting` state in a controlling browsing context.

Presumably if a connection has not been established to the presentation screen, then the connection can be aborted.  If the connection has been established and the presentation started, but the `connected` state change has not yet occurred, that connection can be used to abort or terminate the presentation.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/presentation-api/issue-421-terminate-reconnect.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/presentation-api/38f2399...e2c6d68.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2589,8 +2589,9 @@
           <ol>
             <li>If the <a>presentation connection state</a> of
             <var>connection</var> is not <a for="PresentationConnectionState">
-              connected</a> or <a for="PresentationConnectionState">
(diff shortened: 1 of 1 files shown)


tidoust:
Proposed changes look good to me.

Issue #345, opened 2016-09-16
Introduction could use some copy editing


mfoltzgoogle:
The introduction could use a copy editing pass to improve phrasing and consistency.  This issue is an Editor's TODO to himself :)

PR #429, opened 2017-05-04
Copy editing of non-normative text

mfoltzgoogle:
Addresses Issue #345: Introduction could use come copy editing.

This updates the prose in the introduction to use more precise language and improve style.
It makes minor editing improvements to non-normative notes and the security and privacy section.
It also adds acknowledgements of registered trademarks used in the specification.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/presentation-api/issue-345-editing.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/presentation-api/6c3a68f...fe784b4.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -3318,7 +3318,9 @@
         <em>AirPlay</em>, <em>HDMI</em>, <em>Chromecast</em>, <em>DLNA</em> and
         <em>Miracast</em> are registered trademarks of Apple Inc., HDMI
         Licensing LLC., Google Inc., the Digital Living Network Alliance, and
-        the Wi-Fi Alliance, respectively.
+        the Wi-Fi Alliance, respectively.  They are only cited as background
+        information and their use is not required to implement the
+        specification.
       </p>
     </section>
     <section class="appendix">
--- a/index.html
+++ b/index.html
@@ -138,8 +138,8 @@
   <body>
     <section id="abstract">
       <p>
-        This specification defines an API to enable web content to access
-        external presentation-type displays and use them for presenting web
+        This specification defines an API to enable Web content to access
+        presentation displays and use them for presenting Web
         content.
       </p>
     </section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3308,10 +3312,16 @@
         Kanai, Tobie Langel, Tomoyuki Shimizu, Travis Leithead, and Wayne Carr
         for help with editing, reviews and feedback to this draft.
       </p>
+      <p>
+        <em>AirPlay</em>, <em>HDMI</em>, <em>Chromecast</em>, <em>DLNA</em> and
```
anssiko: @tidoust, is the recommendation to explicitly call these out in the spec, or does the "Non-W3C Trademarks; Member Trademarks" section in https://www.w3.org/Consortium/Legal/2002/trademarks-20021231 cover specs as well? I assume the term "Site" in this generic boilerplate refers to any resource hosted on the w3.org domain, including specs in the TR space.

index.html
```diff
@@ -176,58 +175,62 @@
...
-        attaching to displays could be enhanced in the future by defining a
-        standard protocol for delivering these types of messages that display
-        devices could choose to implement.
+        For example, if the <a>presentation display</a> uses HDMI or Miracast,
+        the user agent (UA) hosting the <a>controller</a> will also render the
+        <a href="#dfn-receiving-browsing-context">presentation</a>. It then
+        uses the operating system to send the resulting graphical and audio
+        output to the presentation display. We refer to this situation as the
+        <dfn data-lt="1-UA">1-UA mode</dfn> implementation of the Presentation
+        API. The only requirement is that the user agent be able to send
```
anssiko: s/be able to/is able to/

anssiko:
Thanks @mfoltzgoogle, great work. Before landing, I'd like @tidoust to clarify how we should handle the non-W3C trademarks.

mfoltzgoogle:
I made another pass to polish the introduction and the Cross Origin Access section.  I'm pretty happy now :)

To improve legibility, I updated to link the first time a term like _presentation_ appears in a given paragraph, and not link following occurrences.

index.html
tidoust: I checked internally. Specs we publish usually do not contain non-W3C trademarks for various reasons. In our case, mentioning these technologies as examples in the introduction is important to clarify the context for readers, and the introduction is clear that other technologies are available.

It is a good idea to call these trademarks out explicitly in the acknowledgement section. The proposed text is fine. I recommend adding "They are only cited as background information" to the end of it to make it crystal clear that one does not need them to implement the specification.

tidoust:
@mfoltzgoogle, @anssiko, see minor proposed tweak to ack section inline. That looks good to me otherwise.

index.html

Issue #455, opened 2018-06-19
Clarify behavior when the receiving page attempts to navigate itself


tidoust:
Issue initially raised in https://github.com/w3c/presentation-api/issues/449#issuecomment-397597655

In [6.6.1](https://w3c.github.io/presentation-api/#creating-a-receiving-browsing-context), the Presentation API spec says that, in receiving contexts, "The top-level browsing context MUST NOT be allowed to navigate itself", but it does not explain what is to happen if it attempts to do so.

Two main possibilities:
1. Abort with a `SecurityError`.
2. Close the receiving browsing context.

We should probably be doing 1. (see https://github.com/w3c/presentation-api/issues/449#issuecomment-397771514): it is consistent with the [navigate](https://www.w3.org/TR/html53/browsers.html#navigated) algorithm in HTML when the context is not "allowed to navigate" and is developer-friendly since it allows them to debug the situation. Problem is it is harder to implement and it does not match what Chrome is currently doing. Second possibility matches how Chrome currently behaves, is easier to implement, but makes it harder for developers to debug and is not consistent with the usual navigation algorithm.

mfoltzgoogle:
Spec comments:
- Our spec should reference the _allowed to navigate_ algorithm in HTML because we already referencing the _sandboxed top-level navigation browsing context flag_ that is used by it. 
- However presentation receivers also block self-navigation, which has no equivalent in HTML.  The spec should be clear this is an additional condition on top of HTML _allowed to navigate_.

Implementation comments:
- Because there is no flag in HTML for the self-navigation case, aborting navigation or throwing is more challenging.
- I don't see a strong use case for receiving user agents offering to open in a new window, but no reason to block it.
- Therefore, in the spec I recommend that the behavior for blocked navigation SHOULD be consistent with HTML.

I can put together a PR along these lines for review.

In the long run, this requirement may be too restrictive.  For example, in Chrome PWA's are allowed to self-navigate within their URL scope without showing an address bar.   We should think about the tradeoffs here.

mfoltzgoogle:
Regarding Chrome's behavior specifically, filed https://crbug.com/949350 to improve the behavior here, but it's tricky.

mfoltzgoogle:
Closed by PR #461.

PR #461, opened 2019-04-04
Clarify self-navigation behavior in a presentation.

mfoltzgoogle:
This addresses feedback in Issue #455, to link navigation behavior to the HTML 5.1 navigation steps, and clarify how the receiving user agent should respond.

There's also a bunch of changes from attempting to run tidy-html5 which I'll clean up in another commit on this PR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/pull/461.html" title="Last updated on Apr 5, 2019, 6:16 PM UTC (45813d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/461/ff5a520...45813d3.html" title="Last updated on Apr 5, 2019, 6:16 PM UTC (45813d3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US">
+<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml" xml:lang="en-US">
   <head>
     <meta charset="UTF-8" />
     <title>
@@ -8,6 +8,7 @@
     <script src="https://www.w3.org/Tools/respec/respec-w3c-common" async=
     "async" class="remove"></script>
     <script class="remove">
+    <![CDATA[
     var respecConfig = {
         specStatus: 'ED',
         edDraftURI: 'https://w3c.github.io/presentation-api/',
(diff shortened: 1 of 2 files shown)


mfoltzgoogle:
PTAL @tidoust as you filed the original issue :-)

index.html
```diff
@@ -3057,14 +3062,26 @@ <h4>
...
             presentation URL shown when <a data-lt=
             "select a presentation display">selecting a presentation
             display</a>.
           </p>
+          <p>
+            If the top-level-browsing context was not <a>allowed to
+            navigate</a>, it SHOULD NOT offer to open the resource in a new
+            <a>top-level browsing context</a>, but instead SHOULD abort the
+            navigation and possibly throw a <a>SecurityError</a> (consistent
+            with the <a>steps to navigate</a>).
```
tidoust: That is not entirely "consistent" with the steps to navigate: the steps to navigate in HTML mandate the `SecurityError` be thrown when the [exceptions enabled flag](https://www.w3.org/TR/html51/browsers.html#exceptions-enabled-flag) is set, whereas this makes it optional. 

If the `SecurityError` exception cannot be mandated because it remains a challenge implementation-wise, I would suggest to drop "(consistent with the steps to navigate)".

mfoltzgoogle: Throwing the exception isn't an implementation concern from my point of view.  Rather I didn't want to get into the details here of when to throw a `SecurityError` and thus duplicate HTML.  I rephrased this, let me know if that addresses your feedback.

tidoust: Yes, looks good!

mfoltzgoogle:
Thanks!

Issue #476, opened 2020-03-03
Same-screen presentation


dlibby-:
I was curious if there has been any discussion of a use case for using the PresentationAPI to present to the same screen where the PresentationRequest originates. I tried searching through the issues, but was unable to find previous discussion.

PowerPoint online is evaluating using the PresentationAPI, and were interested in supporting slideshow 'presentation' (i.e. rehearsing) even when in single screen mode. They would also like to have a single unified codepath for handling this, rather than falling back to window.open/requestFullscreen. Additionally, cases such as mirrored single displays could also use this pattern of having one codepath that just uses the PresentationAPI.

As far as I can tell, there is nothing in the spec that disallows for this usage, but at the same time, given the goals of the spec, 'presenting' to the same screen certainly feels unintuitive. As far as I've seen, current implementations only display secondary displays as available choices for presentation requests.

anssiko:
Thanks @dlibby-. That's a really good real-world use case and as you rightly pointed out the Presentation API spec does not disallow such usage: the [presentation display](https://w3c.github.io/presentation-api/#dfn-presentation-display) can be any graphical and/or audio output device available to the user agent.

@mfoltzgoogle I vaguely recall you may have shared a UI mock or a prototype of such a picture-in-picture usage of Presentation API at some point, but I was unable to find a reference?

This usage would be similar to what [Picture-in-Picture](https://w3c.github.io/picture-in-picture/) is for [Remote Playback API](https://w3c.github.io/remote-playback/), but for presenting any web content and not limited to video elements.

dlibby-:
Thanks @anssiko - having a Picture-in-Picture surface sounds like a very interesting idea! However, I probably wasn't clear enough in my initial post as to what PowerPoint is trying to accomplish.

Today in PowerPoint's desktop application, if you click on the 'slide show' button in the bottom right corner, you will see the slides in fullscreen. If there is a second screen, the slides will be fullscreen on the second screen, and the original view will be changed to a notes view. If there is only one screen, the slides show up as fullscreen on the one screen, and the original view is not changed at all (though it is not visible unless you, e.g. alt-tab to switch back to it).

They are interested in supporting the latter case via PresentationAPI, but currently no user agent that I'm aware of will provide a choice of the current screen.

[...]

PR #479, opened 2020-05-08
Add support for allowing same-screen display

dlibby-:
Addresses #476.
Add an options dictionary to PresentationRequest.start() that indicates the list of available displays that the user can choose from must include the current display.

I thought about plumbing this through to "monitor the list of available presentation displays", but this won't ever be used for getAvailability() and it wasn't clear to me how to mechanically do this in the spec (all ears if there are suggestions though!).

Once there's some consensus that this is the right approach, I'll work on adding a WPT and reference it in this PR description.

***
<a href="https://pr-preview.s3.amazonaws.com/dlibby-/presentation-api/pull/479.html" title="Last updated on May 8, 2020, 7:22 PM UTC (54013d8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/479/1d9fcf3...dlibby-:54013d8.html" title="Last updated on May 8, 2020, 7:22 PM UTC (54013d8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1224,11 +1224,15 @@
           Interface <dfn>PresentationRequest</dfn>
         </h3>
         <pre class="idl">
+          dictionary PresentationRequestStartOptions {
+            boolean allowSameScreen = false;
+          };
+
           [SecureContext, Exposed=Window]
           interface PresentationRequest : EventTarget {
             constructor(USVString url);
             constructor(sequence&lt;USVString&gt; urls);
-            Promise&lt;PresentationConnection&gt; start();
+            Promise&lt;PresentationConnection&gt; start(optional PresentationRequestStartOptions options = {} );
             Promise&lt;PresentationConnection&gt; reconnect(USVString presentationId);
             Promise&lt;PresentationAvailability&gt; getAvailability();
 
@@ -1334,6 +1338,11 @@
               object that received the call to <a data-link-for=
               "PresentationRequest">start</a>
             </dd>
+            <dd>
+              <var>options</var>, the <a>PresentationRequestStartOptions</a>
+              dictionary passed to <a data-link-for=
+              "PresentationRequest">start</a>
+            </dd>
             <dt>
               Output
             </dt>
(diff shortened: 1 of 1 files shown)


anssiko:
Thanks! @mfoltzgoogle to advise on your second option to plumb this through https://w3c.github.io/presentation-api/#monitoring-the-list-of-available-presentation-displays

dlibby-:
@mfoltzgoogle - can you take a look when you get a chance?

takumif:
FYI Mark is unavailable until next Monday.

mfoltzgoogle:
@dlibby- Sorry for the very delayed response, I was unable to spend much time on SSWG issues until recently.

If I understand correctly, the desire is to use the Presentation API to open a new browser window and have it possibly shown on the current display, or a local secondary display or remote display (which are all the possible scenarios I can think of).

If the user chooses the current display, were you assuming that the window with the presentation will be shown fullscreen on that display?

Is it important that these possibilities are all shown in the same dialog?  One solution that is possible without adding to the Presentation API is to use `window.open()` followed by [`document.requestFullscreen()`](https://fullscreen.spec.whatwg.org/#dom-element-requestfullscreen)

There's also the [Window Placement API](https://github.com/webscreens/window-placement/blob/master/EXPLAINER.md#support-requests-to-show-elements-fullscreen-on-a-specific-screen) in incubation, which would allow you to place a fullscreen document on the current display or an external display.  That may be a better fit depending on your use case.

dlibby-:
Yes, your understanding is correct.

I don't think `window.open()` + `requestFullscreen()` would typically work as user activation is not preserved across `window.open()` calls (it's been a while since I tested this so perhaps this has changed, or is open for debate).

Window Placement looks like it would suffice in terms of being able to achieve the end result. Prior to that, I don't believe there was any way to detect this particular scenario.

I believe one goal that wouldn't be satisfied by Window Placement would be that the webpage must maintain separate codepaths. Whereas a window.open solution would probably postMessage between opener/opened window, the flow for successfully using Presentation API would use PresentationConnection.send. Similarly, monitoring disconnection/window closing would require different code.

anssiko:
@dlibby- closed #476 a while ago so I'll close this PR as well.

Issue #202, opened 2015-10-06
Presentations without communication channel


tidoust:
The Presentation API assumes that the user agent handles the discovery, control and the establishment of the communication channel. In issue #61 (Add facility for the opening page to add cloud paired screens as presentation targets), we have already started to discuss how the Web app could perhaps take care of all these steps and register online screens under its control with the user agent, so that the user be prompted only once.

I wonder whether something in-between could be useful as well, in other words a way to let the user agent handle the discovery and control, while the Web app takes care of the communication channel, typically going through some backend server.

In terms of API, the change I have in mind would be a new `options` parameter for the `PresentationRequest` constructor that Web apps could use to set an `isChannelOptional` flag. When that flag is set, the user agent would not need to establish a communication channel between the contexts and could thus list additional devices or mechanisms that could be used for the presentation. By default, this flag is not set, only displays with which the user agent may establish a communication channel are listed, and Web apps can continue to assume that such a channel will eventually be created. Other changes may perhaps be needed, or this may be doable differently.

Examples of presentation mechanisms that could be implemented if such a flag existed:

1. HbbTV 2.0 (as noted in #67, the receiving Web app will need to establish the communication channel with the local WebSocket server on its own)
2. Other DIAL applications, at least pending a proper mechanism to establish a communication channel
3. QR code. So passé, I know...
4. Broadcasting the URL over a Bluetooth LE beacon, à la Physical Web.
5. Tapping devices using NFC.
6. Using Push notifications on devices on previously paired devices, through a Service Worker.

Some of these mechanisms may be implemented at the app-level (Push API, QR code) and are thus more linked to issue #61. Others may be doable in the future (NFC) or may trigger additional privacy concerns (e.g. broadcasting the URL) as they are not discovery-based but rather invitation-based. In some cases, there is no guarantee that the controlling device will even know that a receiving device has loaded the URL or that only one such device will do so.

However, practically speaking, all of these mechanisms would allow a Web application to present web content on a secondary display.

I experimented with @dontcallmedom within the MediaScape EU project on some of these mechanisms in an updated version of the Presentation API polyfill that I had written for the Slidy Remote demo. The code of this new polyfill is available at:

  https://mediascape.github.io/presentation-api-polyfill/

tidoust:
See relevant discussion at TPAC F2F:
http://www.w3.org/2015/10/28-webscreens-minutes.html#item07

tidoust:
[...]

PR #484, opened 2020-10-15
Add note on how presentations without connections should be handled.

mfoltzgoogle:
Addresses Issue #202: Presentations without communication channel

This advises user agents on how to handle presentations where the controlling page cannot communicate with the presentation through the Presentation API. 

Presumably, the presentation request URL contains sufficient information for the two pages to rendezvous through some other mechanism (if they wish to communicate).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/pull/484.html" title="Last updated on Oct 15, 2020, 12:06 AM UTC (2ec5092)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/484/7cc74aa...2ec5092.html" title="Last updated on Oct 15, 2020, 12:06 AM UTC (2ec5092)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2343,6 +2343,19 @@
             and in-order fashion as described in the <a>Send a Message</a> and
             <a>Receive a Message</a> steps below.
           </div>
+          <p class="note">
+            In some cases, the <a>receiving user agent</a> may not support
+            communication with the <a>controlling user agent</a> through the
+            Presentation API. For example, the <a>controlling browsing
+            context</a> and <a>receiving browsing context</a> could use WebRTC
+            or a cloud service to communicate independently of
+            a <a>PresentationConnection</a>. In these cases, after step 3 above,
+            user agents can immediately close the connection with
+            a <a data-link-for= "PresentationConnectionCloseEvent">reason</a>
+            of <a data-link-for="PresentationConnectionCloseReason">error</a>
+            and a <dfn data-link-for="PresentationConnectionCloseEvent">message
+            </dfn> indicating that connections are not supported.
+          </p>
         </section>
         <section>
           <h4>


mfoltzgoogle:
@tidoust, you had the most to say on this long-standing issue :-)

tidoust:
A couple of comments:

1. Doesn't the note actually contradict the note that's right before it and that says: "The connection *must provide a two-way messaging abstraction* [...]"?
2. In particular, many apps will expect a communication channel to be available. This note suggests that user agents may not provide one. Isn't that going to break the user experience?

In other words, I believe that the controlling user agent should never offer to connect to a receiving user agent that wouldn't be able to provide a communication channel, unless the application is explicit that the communication channel is optional (or unless we're talking about a URI scheme different from `http` and `https`, but then that's out of scope of the spec). That is the reason why I was proposing some sort of option to the `PresentationRequest` constructor.

In any case, if there are no implementation plans for such a feature, it should probably just be dropped.

mfoltzgoogle:
> In other words, I believe that the controlling user agent should never offer to connect to a receiving user agent that wouldn't be able to provide a communication channel, unless the application is explicit that the communication channel is optional (or unless we're talking about a URI scheme different from http and https, but then that's out of scope of the spec). That is the reason why I was proposing some sort of option to the PresentationRequest constructor.

I guess you are right, in that this PR doesn't address the problem of availability - the user agent won't know whether to offer receivers without a communication channel, since it won't know in advance whether the application is capable providing its own.

> In any case, if there are no implementation plans for such a feature, it should probably just be dropped.

We don't have any plans to offer the types of endpoints #3-#6 in Issue #202 (like NFC, BTLE, QR codes) as "presentation" devices through this API.  Since the issue was filed, the are Web APIs exposing these capabilities have matured and I think it would make more sense for developers to try those first.

I'm going to close this PR and the original issue.

Issue #509, opened 2022-12-14
Update spec language to refer to child navigables instead of nested browing contexts?


mfoltzgoogle:
The term `nested browsing context` has apparently been removed in the WhatWG HTML spec and replaced with the concept of a `child navigable` [1, 2].  I believe we can do a 1-1 replacement of the terms without changing the meaning of anything, but need to check.

Right now, ReSpec is complaining that it can't find an x-ref for `nested browsing context` as that term no longer exists in WhatWG HTML.

[1] https://github.com/whatwg/html/pull/6315
[2] https://html.spec.whatwg.org/multipage/document-sequences.html#child-navigable

PR #514, opened 2023-01-18
Update cross references to HTML.

mfoltzgoogle:
See Issue #509 for background.   This PR is intended to fix the current crop of ReSpec errors and warnings.  

WhatWG HTML makes the following recommendation,

> Modern specifications should avoid using the browsing context concept in most cases, unless they are dealing with the subtleties of browsing context group switches and agent cluster allocation. Instead, the Document and navigable concepts are usually more appropriate.

In the future we may want to update the spec to refer to current concepts like navigables and traversables, instead of browsing contexts.  However that is out of scope for right now.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/pull/514.html" title="Last updated on Jan 19, 2023, 2:48 PM UTC (e7f90b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/514/cb625d4...e7f90b9.html" title="Last updated on Jan 19, 2023, 2:48 PM UTC (e7f90b9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -853,7 +853,7 @@
         </p>
         <p>
           The <a>task source</a> for the tasks mentioned in this specification
-          is the <dfn>presentation task source</dfn>.
+          is the <var>presentation task source</var>.
         </p>
         <p>
           Unless otherwise specified, the <a>JavaScript realm</a> for script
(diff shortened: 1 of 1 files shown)


tidoust (approved):
I took the liberty to push an update. The "list of descendant browsing contexts" also needed to be replaced by "descendant navigables", which also has the benefit to be exported by the HTML spec.

Also, I think we should keep a link to the notion of "unload a document", even though the HTML spec no longer exports that notion, since it helps clarify what implementations should do.

PR #524, opened 2024-09-15
Fix task queueing.

markafoltz:
Addresses Issue #523 by specifically queuing a global task when promises are resolved or events are fired from parallel steps.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/pull/524.html" title="Last updated on Sep 19, 2024, 6:31 PM UTC (0fb1088)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/presentation-api/524/74d1a41...0fb1088.html" title="Last updated on Sep 19, 2024, 6:31 PM UTC (0fb1088)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1113,9 +1113,8 @@
               </ol>Then run the following steps:
               <ol>
                 <li>
-                  <a data-cite="!HTML#queue-a-global-task">Queue a global
-                  task</a> to <a>reject</a> <var>P</var> with a
-                  {{NotFoundError}} exception.
+                  <a>Queue a global task</a> to <a>reject</a> <var>P</var> with
+                  a {{NotFoundError}} exception.
                 </li>
                 <li>Abort all remaining steps.
                 </li>
@@ -1754,8 +1753,8 @@
                 <var>newAvailability</var> and skip the following step.
                 </li>
                 <li>If <var>previousAvailability</var> is not equal to
-                <var>newAvailability</var>, then <a>queue a task</a> to run the
-                following steps:
+                <var>newAvailability</var>, then <a>queue a global task</a> to
+                run the following steps:
                   <ol>
                     <li>Set <var>A</var>'s <a>value</a> property to
                     <var>newAvailability</var>.
@@ -2048,8 +2047,8 @@
             identifier</a> of <var>presentationConnection</var> MUST be sent
             with this request.
             </li>
-            <li>If connection completes successfully, <a>queue a task</a> to
-            run the following steps:
+            <li>If connection completes successfully, <a>queue a global
+            task</a> to run the following steps:
               <ol>
                 <li>Set the <a>presentation connection state</a> of
                 <var>presentationConnection</var> to <a data-link-for=
@@ -2466,7 +2465,7 @@
                 connection</var> is <a data-link-for=
                 "PresentationConnectionState">connected</a> or
                   <a data-link-for="PresentationConnectionState">connecting</a>,
-                  then <a>queue a task</a> to run the following steps:
+                  then <a>queue a global task</a> to run the following steps:
                   <ol>
                     <li>Set the <a>presentation connection state</a> of
                     <var>known connection</var> to <a data-link-for=
@@ -2548,8 +2547,8 @@
             the user interface and discard it.
             </li>
             <li>For each <var>connection</var> in
-            <var>connectedControllers</var>, <a>queue a task</a> to send a
-            termination confirmation for <var>P</var> using an implementation
+            <var>connectedControllers</var>, <a>queue a global task</a> to send
+            a termination confirmation for <var>P</var> using an implementation
             specific mechanism to the <a>controlling user agent</a> that owns
             the <a>destination browsing context</a> for <var>connection</var>.
               <p class="note">
@@ -2572,7 +2571,7 @@
           <ol>
             <li>For each <var>connection</var> in the <a>set of controlled
             presentations</a> that was connected to <var>P</var>, <a>queue a
-            task</a> to run the following steps:
+            global task</a> to run the following steps:
               <ol>
                 <li>If the <a>presentation connection state</a> of
                 <var>connection</var> is not <a data-link-for=
--- a/index.html
+++ b/index.html
@@ -2587,7 +2587,9 @@
           <ol>
             <li>For each <var>connection</var> in the <a>set of controlled
             presentations</a> that was connected to <var>P</var>, <a>queue a
-            Presentation API task</a> to run the following steps:
+            global task</a> on the <a>presentation task source</a> given <var>
+              connection</var>'s <a>relevant global object</a> to run the
+              following steps:
               <ol>
                 <li>If the <a>presentation connection state</a> of
                 <var>connection</var> is not <a data-link-for=
--- a/index.html
+++ b/index.html
@@ -49,7 +49,7 @@
         sotdAfterWGinfo: true,
         group: 'secondscreen',
         github: 'https://github.com/w3c/presentation-api',
-        xref: ['dom', 'fileapi', 'secure-contexts', 'html', 'url', 'webidl', 'webrtc', 'websockets'],
+        xref: ['dom', 'fileapi', 'secure-contexts', 'html', 'url', 'webidl', 'webrtc', 'websockets', 'infra'],
         localBiblio: {
           DIAL: {
             title: 'DIscovery And Launch Protocol Specification',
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -840,6 +841,10 @@ <h3>
         <p>
           Unless otherwise specified, the <a>JavaScript realm</a> for script
           objects constructed by algorithm steps is the <a>current realm</a>.
+          Unless otherwise specified, all global tasks use the <a data-cite=
+          "!HTML#networking-task-source">networking task source</a> associated
+          with the <a data-cite="!HTML#global-object">global object</a> of the
+          <a>current realm</a>.
```
tidoust, line 847: Use of the "networking task source" contradicts the previous paragraph, which defines a dedicated "presentation task source".

I think it is good practice to create a dedicated task source unless there's a risk that tasks conflict with other tasks defined in other specs. I don't see any reason why Presentation API tasks would conflict with other networking tasks and require relative ordering with them.

index.html
```diff
@@ -1037,6 +1042,10 @@ <h4>
           <h4>
             Selecting a presentation display
           </h4>
+          <p>
+            The global tasks queued in these steps use the <a data-cite=
+            "!HTML#-interaction-task-source">user interaction task source</a>.
+          </p>
```
tidoust, line 1048: Same as above, the "user interaction task source" is used in HTML for tasks that fire events in direct response to user interaction. Even though transient activation is required, the call to `start()` is programmatic.

tidoust (changes requested):
Part of the problem with missing tasks is indeed figuring out what task source to use. If there are no good reasons to reuse existing ones, the best is to use a dedicated one.

The tool that analyzed the spec and reported #523 is not smart enough to follow algorithms and/or understand which algorithms de facto run in parallel. If it were, it would also have reported the [start a presentation connection](https://w3c.github.io/presentation-api/#dfn-start-a-presentation-connection) algorithm. The algorithm also needs fixing.

Ideally, that algorithm would start with a step that says "[Assert](https://infra.spec.whatwg.org/#assert): this is running [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel)." step, as done in a few algorithms, e.g., [attempt to populate the history entry's document in HTML](https://html.spec.whatwg.org/multipage/browsing-the-web.html#attempt-to-populate-the-history-entry's-document), so that the tool could understand that it actually runs in parallel.

Speaking of the analysis tool, I understand why it's convenient to proceed as you did, but having a separate "all tasks use the task source 'foo' and the global object of the current realm" in prose is the kind of things that makes it hard to analyze the spec semi-automatically. I started a [thread on spec-prod](https://lists.w3.org/Archives/Public/spec-prod/2024JulSep/0006.html) partly to discuss conventions that could strike a good balance between readability, ease of authoring, and ease of validation.

One possible alternative to avoid inlining the selection of the global object and repeating the name of the task source as done in a few other specs, while still making it possible to analyze the spec would be to define a "Queue a Presentation API task" algorithm that initializes the variables and calls "Queue a global task", and to rather call that algorithm within the other algorithms. The analysis tool will miss that today, but could at least in theory process that somewhat easily afterwards. Just a suggestion, it's ok to keep the definition in prose.

index.html
markafoltz: I forgot that the spec defines its own presentation task source.  Will remove the reference to the networking task source.

index.html
markafoltz: I used the "user interaction task source" since the user agent should (in nearly all cases) show some sort of user interface.  But it's definitely not a requirement.  So I will remove the reference here.

markafoltz:
> define a "Queue a Presentation API task" algorithm that initializes the variables and calls "Queue a global task"

I went ahead and made this change.   I also added an "in parallel" assertion for steps that require the Presentation API task source, without explicitly mentioning running "in parallel."

tidoust (commented):
The asserts makes me realize that the introduction of the algorithm to [start a presentation from a default presentation request](https://w3c.github.io/presentation-api/#starting-a-presentation-from-a-default-presentation-request) could say that the user agent must run the algorithm "in parallel" too (or another assert could be added).

Also, it's not clear to me that the algorithm that [terminates a connection in a controlling browsing context](https://w3c.github.io/presentation-api/#terminating-a-presentation-in-a-controlling-browsing-context) actually runs in parallel because it is called directly when the application calls `terminate()`. If it needs to run in parallel, the call to it should probably indicate that the user agent must run the algorithm in parallel.

index.html
```diff
@@ -2035,8 +2059,8 @@ <h4>
             identifier</a> of <var>presentationConnection</var> MUST be sent
             with this request.
             </li>
-            <li>If connection completes successfully, <a>queue a task</a> to
-            run the following steps:
+            <li>If connection completes successfully, <a>queue a global
+            task</a> to run the following steps:
```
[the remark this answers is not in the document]
markafoltz: Done

index.html
```diff
@@ -2934,8 +2971,8 @@ <h4>
                 the <a>set of presentation controllers</a>.
                 </li>
                 <li>
-                  <a>Queue a task</a> to [=fire an event=] named
-                  <a data-link-for=
+                  <a>queue a Presentation API task</a> to [=fire an event=]
```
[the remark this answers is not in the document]
markafoltz: Done

index.html
```diff
@@ -835,7 +836,14 @@ <h3>
...
         <p>
           The <a>task source</a> for the tasks mentioned in this specification
-          is the <var>presentation task source</var>.
+          is the <dfn>presentation task source</dfn>.
+        </p>
+        <p>
+          When an algorithm <dfn data-lt="queue a Presentation API task">queues
+          a Presentation API task</dfn> <var>T</var>, the <a>user agent</a>
+          MUST <a>queue a global task</a> <var>T</var> on the <a>presentation
+          task source</a> using the <a data-cite="!HTML#global-object">global
```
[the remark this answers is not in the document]
markafoltz: Done

index.html
```diff
@@ -1221,6 +1229,10 @@ <h4>
             </dd>
           </dl>
           <ol>
+            <li>
+              <a data-cite="!INFRA#assert">Assert</a>: this is running <a>in
```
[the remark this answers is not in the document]
markafoltz: Done

markafoltz:
Thanks for your detailed review and clear comments as usual @tidoust 😺 

> the algorithm to [start a presentation from a default presentation request](https://w3c.github.io/presentation-api/#starting-a-presentation-from-a-default-presentation-request) could say that the user agent must run the algorithm "in parallel" too

Added an explicit step to run these steps in parallel.

> Also, it's not clear to me that the algorithm that [terminates a connection in a controlling browsing context](https://w3c.github.io/presentation-api/#terminating-a-presentation-in-a-controlling-browsing-context) actually runs in parallel because it is called directly when the application calls terminate(). If it needs to run in parallel, the call to it should probably indicate that the user agent must run the algorithm in parallel.

I think the only step that needs to run "in parallel" is notifying the receiving browsing context of the termination.  I updated the steps to clarify that.

index.html
```diff
@@ -2437,6 +2463,9 @@ <h4>
             MUST run the following steps:
           </p>
           <ol>
+            <li>
+              <a>Assert</a>: this is running <a>in parallel</a>.
+            </li>
```
tidoust, line 2468: Given your last comment and update, this algorithm does not run in parallel, only specific steps do.

```suggestion
```

index.html
```diff
@@ -2453,7 +2482,8 @@ <h4>
                 connection</var> is <a data-link-for=
                 "PresentationConnectionState">connected</a> or
                   <a data-link-for="PresentationConnectionState">connecting</a>,
-                  then <a>queue a task</a> to run the following steps:
+                  then <a>queue a Presentation API task</a> to run the
```
tidoust, line 2485: Re-reading this, I'm wondering whether, here and in the "Handling a termination confirmation in a controlling user agent" algorithm, it could be worth expanding that "current realm" means the [relevant realm](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global) of *known connection* (that realm may not be the same for each of them in particular, and may not be the realm under which the algorithm runs).

One possibility could be to rewrite as:
```
<a>queue a global task</a> on the <a>presentation task source</a>
given <var>known connection</var>'s <a>relevant global object</a> to...
```

tidoust (commented):
See inline. One assert to remove, and one additional suggestion to clarify what realm to use in steps that loop through the set of controlled presentations.

index.html
markafoltz: Done.

index.html
markafoltz: That's correct, _known connection_ likely belongs to a different document and thus a different global object.

index.html
```diff
@@ -2559,7 +2587,7 @@ <h4>
           <ol>
             <li>For each <var>connection</var> in the <a>set of controlled
             presentations</a> that was connected to <var>P</var>, <a>queue a
-            task</a> to run the following steps:
+            Presentation API task</a> to run the following steps:
```
tidoust, line 2590: Same update as that done in "Terminating a presentation in a controlling browsing context" algorithm:

```suggestion
            global task</a> on the <a>presentation task source</a> given
            <var>connection</var>'s <a>relevant global object</a> to run the
            following steps:
```

markafoltz: Done

